PR #7, opened 2023-03-30
update example ModelRunner to fetch CT input data

LennyN95:
Our demo models should use semantics to show the purpose better. Interface and expected behaviour don't change, pipeline tested locally.


LennyN95:
Some api changed, e.g., instead of using `instance.getData(...)`, which was built on a filter implemented in the Instance class, you now use `instance.data.filter(...).first()` as `instance.data` is now an `InstanceDataCollection` object instead of a `list[InstanceData]`.

PR #18, opened 2023-05-31
BUG: make values assigned to BodyPartExamined compliant with DICOM

fedorov:
BodyPartExamined content should be populated from the list of defined terms available in this table in the standard:

https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_L.html#chapter_L

Capitalization is important, since BodyPartExamined Value Representation (VR) is Code String (CS) (https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html), which does not allow lower case characters.


LennyN95 (approved):
Good catch!

PR #23, opened 2023-06-14
MHub / GC - WIP: Adding PanImg dataconverter classes for Xie lobe segmentation

silvandeleemput:
### PR changes:

* I have created two dataconverter classes: MhaPanImgConverter (Dicom/NRRD -> MHA), and TiffPanImgConverter (WSI-Dicom -> Tiff), for now these are placed under: `/models/xie2020_lobe_segmentation/utils/PanImgConverters.py`, but are intended to be moved to `mhubio/modules/convert`
* Created a working example by modifing the Xie lobe segmentation algorithm to use the new converters;
  *  Dockerfile: added the required panimg dependency install (will require an update of the MHubAI/models.git commit hash to work correctly, which I wasn't able to include yet!), also should probably be moved as part of the base MHub Docker images at some point after the data converter classes are migrated to mhubio.
  * config.yml: removed the NRRDConverter step, changed the source seg type for the output to MHA
  * run scripts: Added the MHAPanImgConverter to the pipeline
  * LobeSegmentationRunner: Changed the expected IO decorators to MHA

### Comments

I think that this PR should provide a good starting point for properly integrating the panimg conversion tools and for establishing a good standard template for further algorithm exchange. On my machine, I can get the example pipeline to run using the added MhaPanImgConverter, instead of using the NrrdConverter. The input and output types have been changed to MHA, since this will be the default for most of the GC algorithms. This change might break the slicer integration since the output isn't converted back to nrrd at the end of the pipeline, if this is the case we could easily add a converter for that as well. 

As a final minor concern, with my current working version, I am constantly running into the following error, at the end of the pipeline, but I think this should be easily solvable:
```
Start DsegConverter
ERROR: DsegConverter failed processing instance <I:/app/data/sorted_data/1.3.6.1.4.1.14519.5.2.1.6279.6001.179049373636438705059720603192>: __init__() got an unexpected keyword argument 'body_part_examined' in Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/mhubio/core/IO.py", line 102, in wrapper
    func(self, instance, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/mhubio/core/IO.py", line 129, in wrapper
    func(self, instance, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/mhubio/core/IO.py", line 117, in wrapper
    func(self, instance, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/mhubio/core/IO.py", line 165, in wrapper
    func(self, instance, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/mhubio/modules/convert/DsegConverter.py", line 55, in task
    generator = DcmqiDsegConfigGenerator(
TypeError: __init__() got an unexpected keyword argument 'body_part_examined'
```

This error seems to be related to a version mismatch in the DcmqiDsegConfigGenerator __init__ interface not yet expecting the `body_part_examined` argument and seems part of the latest MHub base image that I use. What would be the recommended way to solve this?

### DIFF
--- a/models/xie2020_lobe_segmentation/config/config.yml
+++ b/models/xie2020_lobe_segmentation/config/config.yml
@@ -2,7 +2,6 @@ general:
   data_base_dir: /app/data
   
 modules:
-
   DicomImporter:
     source_dir: input_data
     import_dir: sorted_data
@@ -15,3 +14,5 @@ modules:
   DsegConverter:
     dicomseg_json_path: /app/models/xie2020_lobe_segmentation/config/dseg.json
     skip_empty_slices: True
+    body_part_examined: LUNG
+    source_segs: ['mha:mod=seg:roi=*']
--- a/models/xie2020_lobe_segmentation/dockerfiles/cuda11.4/Dockerfile
+++ b/models/xie2020_lobe_segmentation/dockerfiles/cuda11.4/Dockerfile
@@ -9,6 +9,12 @@ RUN pip3 install --no-cache-dir \
     dgl-cu113 -f https://data.dgl.ai/wheels/repo.html \
     torch==1.10.0+cu113 torchvision==0.11.1+cu113 -f https://download.pytorch.org/whl/cu113/torch_stable.html
 
+# Install panimg to test conversion integration TODO should later be installed with MHub/mhubio
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  python3-openslide \
+  && rm -rf /var/lib/apt/lists/*
+RUN pip3 install panimg
+
 # SimpleITK downgrade required for legacy Resample::Execute operation
 RUN pip3 install --no-cache-dir --force-reinstall SimpleITK==1.2.4
 
--- a/models/xie2020_lobe_segmentation/dockerfiles/cuda11.4/Dockerfile
+++ b/models/xie2020_lobe_segmentation/dockerfiles/cuda11.4/Dockerfile
@@ -2,7 +2,7 @@
 FROM mhubai/base:cuda11.4
 
 # Specify/override authors label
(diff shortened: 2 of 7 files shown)


LennyN95:
Hi Sil,

things moved fast during PW and run out of sync with mhubio and segdb repositories - as you've noticed. In this case, using the latest build of dockerhub instead of rebuilding the image should work. By now, the repositories should be in sync.

LennyN95:
I looked into the code in the meantime.

```
for result in img_builder(files=input_files):
    sitk_image = result.image  # SimpleITK image
    SimpleITK.WriteImage(sitk_image, str(out_mha_file))
```

Can you explain how, e.g., the dicom image builder works? As far as I can see, it receives any number of `*.dcm` files and then returns one image per study? If the above example iterates more than once, we'd override `out_mha_file`. So I guess we expect exactly one iteration for the Dicom data of a single CT scan - right?

silvandeleemput:
> Can you explain how, e.g., the dicom image builder works? As far as I can see, it receives any number of `*.dcm` files and then returns one image per study? If the above example iterates more than once, we'd override `out_mha_file`. So I guess we expect exactly one iteration for the Dicom data of a single CT scan - right?

I think you are correct [see comments/code here](https://github.com/DIAGNijmegen/rse-panimg/blob/2c83742722439f66569c3b4bd553fd2e06e4472d/panimg/image_builders/dicom.py#L407-L411), it should process the input *.dcm (or *.nrrd) files and output these per series instance UID (in the general case), overwriting the `out_mha_file` location if multiple are encountered. Might be cleaner/better to output these separately though or break the code after the first instance, could even be an importer option.

PR #27, opened 2023-07-04
MHub / GC - Add grt123 Model for lung cancer prediction based on lung nodules

silvandeleemput:
Hi, this PR contains the required code for getting our modified version of the publicly available [grt123 model](https://github.com/DIAGNijmegen/bodyct-dsb2017-grt123), winner of the Data Science Bowl 2017 Kaggle challenge. 

### Caveats
* The PR target is `main`, but should be something like `m-gc-grt123-lung-cancer`
* In the Dockerfile the MHub model integration is currently marked `TODO` since it requires the creation of the appropriate branch for this code first.
* This implementation still uses a run.py script because the MHAConverter doesn't have the panimg backend yet.

### Algorithm I/O
  * Input should be a CT Lung image (MHA) (This MHub implementation expects a Dicom which is converted to MHA using MhaPanImgConverter)
  * Output is a JSON file containing the predicted nodule findings (locations) and predicted cancer probability scores.
  * Test images for this algorithm can be found [here](https://github.com/DIAGNijmegen/bodyct-dsb2017-grt123/blob/9a4ca0415c7fc1d3023a16650bf1cdce86f8bb59/tests/resources/inputs.tar.xz)


silvandeleemput:
This PR has been updated for the new base image.

LennyN95:
Some suggestions
- For the final prediction score, add a `ValueOutput` to the Runner module
- Remove the`tmp_path` config
- Use `requestTempDir` for temporary folders ([example](https://github.com/MHubAI/models/blob/fb4d9897b4d392bb4b85799efe9d3a3506078e63/models/casust/utils/CasustRunner.py#L35))
- Remove custom gpu checks if not vital for the model runner *(we may implement such a feature globally in mhubio, let's discuss this in our next meeting!)*

silvandeleemput:
I just updated and cleaned up this PR:
* default.yml
  * renamed from config.yml
  * added version and description
  * updated pipeline with panimg mhaconverter
* Dockerfile
  * added fixed commit hash for grt123 repo git clone
  * updated entrypoint
* LungCancerClassifierRunner.py
  * removed tmp_path config option
  * added requestTempDir for tmp_path
  * added more comments
* Removed script files and custom PanImgConverter

This satisfies all but two of your suggestions:

> * Remove custom gpu checks if not vital for the model runner _(we may implement such a feature globally in mhubio, let's discuss this in our next meeting!)_

Is required for the model to know how the number of GPUs it has available, so I left it in.

> * For the final prediction score, add a `ValueOutput` to the Runner module

This is tricky, since it outputs quite a detailed report with multiple scores per nodule per scan (see below). Maybe it is better to leave it as is? Or maybe we could only add a single cancer probability ValueOutput for the whole scan (last probability value in the report after cancerinfo)).

```json
{
    "lungcad": {
        "revision": "9a4ca0415c7fc1d3023a16650bf1cdce86f8bb59",
        "name": "grt123",
        "datetimeofexecution": "09/13/2023 12:28:46",
        "coordinatesystem": "World",
        "computationtimeinseconds": 48.483961
    },
    "imageinfo": {
        "dimensions": [
            512,
            512,
            140
        ],
        "voxelsize": [
            0.820312,
            0.820312,
            2.5
        ],
        "origin": [
            -228.800003,
            -210.0,
            -379.0
        ],
        "orientation": [
            1.0,
            0.0,
            0.0,
            0.0,
            1.0,
            0.0,
            0.0,
            0.0,
            1.0
        ],
        "seriesuid": "dicom"
    },
    "findings": [
        {
            "id": 0,
            "x": -46.800003000000004,
            "y": -31.0,
            "z": -168.0,
            "probability": 0.9999904632568359,
            "cancerprobability": 0.75371253490448
        },
        {
            "id": 1,
            "x": 73.199997,
            "y": 79.0,
            "z": -191.0,
            "probability": 0.999943733215332,
            "cancerprobability": 0.6662029027938843
        },
        {
            "id": 2,
            "x": 24.199996999999982,
            "y": -47.0,
            "z": -171.0,
            "probability": 0.9999247789382935,
            "cancerprobability": 0.47204485535621643
        },
        {
            "id": 3,
            "x": -82.800003,
            "y": 108.0,
            "z": -159.0,
            "probability": 0.8182298541069031,
            "cancerprobability": 0.004258527886122465
        },
        {
            "id": 4,
            "x": 49.199996999999996,
            "y": -67.0,
            "z": -275.0,
            "probability": 0.7932767271995544,
            "cancerprobability": 0.014665956608951092
        },
        {
            "id": 5,
            "x": 69.19999700000001,
            "y": -63.0,
            "z": -287.0,
            "probability": 0.6662660241127014,
            "cancerprobability": 0.0054708984680473804
        },
        {
            "id": 6,
            "x": 126.199997,
            "y": 28.0,
            "z": -334.0,
            "probability": 0.6302552819252014,
            "cancerprobability": 0.008180161006748676
        },
        {
            "id": 7,
            "x": -62.800003000000004,
            "y": -4.0,
            "z": -102.99999999999999,
            "probability": 0.48750755190849304,
            "cancerprobability": 0.00867788027971983
        },
        {
            "id": 8,
            "x": 73.199997,
            "y": 80.0,
            "z": -215.0,
            "probability": 0.36484286189079285,
            "cancerprobability": 0.03693072870373726
        }
    ],
    "cancerinfo": {
        "casecancerprobability": 0.9574154615402222,
        "referencenoduleids": [
            0,
            1,
            2,
            3,
            4
        ]
    }
}
```

LennyN95:
_**NOTE**: This PR proposes a model currently using a dynamic length output. We are working on a method to export dynamic length outputs and will likely introduce a `@IO.Out.Data.Many` decorator (similar to our `@IO.Output` -> `@IO.Outputs` decorators)._

models/gc_grt123_lung_cancer/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+    rm -rf /gc_grt123_lung_cancer/processor && \
+    rm -rf /gc_grt123_lung_cancer/images && \
+    rm /gc_grt123_lung_cancer/README.md && \
+    rm /gc_grt123_lung_cancer/solution-grt123-team.pdf && \
+    mv /gc_grt123_lung_cancer/.git/HEAD /gc_grt123_lung_cancer && \
+    rm -rf /gc_grt123_lung_cancer/.git/* && \
+    mv /gc_grt123_lung_cancer/HEAD /gc_grt123_lung_cancer/.git
+
+# Clone MHub model (m-gc-grt123-lung-cancer branch, fixed to commit TODO)
+#RUN git init \
```
LennyN95: To avoid disabling the model import during development, please use the recently introduced model import build utility instead:

```
ARG MHUB_MODELS_REPO
RUN buildutils/import_mhub_model.sh gc_grt123_lung_cancer ${MHUB_MODELS_REPO}
```

LennyN95:
_**NOTE**: This PR proposes a model currently using a dynamic length output. We are working on a method to export dynamic length outputs and will likely introduce a `@IO.Out.Data.Many` decorator (similar to our `@IO.Output` -> `@IO.Outputs` decorators)._

LennyN95:
> This is tricky, since it outputs quite a detailed report with multiple scores per nodule per scan (see below). Maybe it is better to leave it as is? Or maybe we could only add a single cancer probability ValueOutput for the whole scan (last probability value in the report after cancerinfo)).

Let's do both, or actually all three :)
- Export the original JSON file
- Use a Value Output for the overall score
- Use a second, dynamic Value Output for all findings.

Dynamic Value Outputs are supported by the newest MHub-IO release now :)
The documentation can be found [here](https://github.com/MHubAI/documentation/blob/main/documentation/mhubio/how_to_write_an_mhubio_module.md#dynamic-output-1n) and are fully supported in our `ReportExporter` Module by using an `aggregate` function (similar to files) but with some additional value operations.

An example implementation would look like this:

```Python
@ValueOutput.Name('lnrisk')
@ValueOutput.Label('Lung Nodule Risk-Score.')
@ValueOutput.Type(int)
@ValueOutput.Description('The predicted risk score for a single lung nodule detected by the alggorithm.')
class LNRisk(ValueOutput):
   pass

def getLungNodulesRiskScores(dicom_dir) -> List[int]:
   # ... find lung nodules, and report back an array of risk scores
   return lst_scores
   
class MyModule(Module):

   @IO.Instance
   @IO.Input('in_data', 'dicom:mod=ct', the='chest CT image')
   @IO.OutputDatas('lnrisks', LNRisk)
   def task(self, instance: Instance, in_data: InstanceData, lnrisks: LNRisk):

      scores = getLungNodulesRiskScores(in_data.abspath)

      for nodule_i, score in enumerate(scores):

         # create value output instance and set the value (we can also modify the description)
         lnrisk = LNRisk()
         lnrisk.description += f" (for nodule {nodule_i})"
         lnrisk.value = score

         # add to collection
         lnrisks.add(lnrisk)
```

models/gc_grt123_lung_cancer/dockerfiles/Dockerfile
silvandeleemput: Done

silvandeleemput:
@LennyN95 I have added the case level score and the dynamic scores per finding. 
I also found that it might be useful for the dynamic scores to set the metadata for associated values (like position and id) per output value, like so:
https://github.com/MHubAI/models/blob/2d4365ad91a33be3bf86f666a1a4768d8e9dbc81/models/gc_grt123_lung_cancer/utils/LungCancerClassifierRunner.py#L114-L119

Setting the metadata in this way helps with debugging as well, with debug output like:
```
├── lncancerprob [Lung Nodule cancer probability score.]
│   The predicted cancer probability score for a single lung nodule detected by the algorithm (for nodule 8 at location (73.199997, 80.0, -215.0))
│   └── Lung Nodule cancer probability score. (0.03693072870373726)
│   ├── id: 8
│   ├── x: 73.199997
│   ├── y: 80.0
│   └── z: -215.0

``` 

Do you think this would be of added value? 
Furthermore, could you give some feedback on the general implementation of the output values, i.e. is it satisfactory?

LennyN95:
Adding the ID (and coordinates) to the metadata is excellent (because then, when exporting the report, you could technically filter by these values). 

However, metadata is used to query files/data and is not available in the report exporter. So if a value is to be exportable, it needs its own value output. We could of course also implement a directive to export data metadata to the report. However, I find that this could create confusion about the ReportExporter, as it is less clear when information is stored in metadata and when as value output.

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -0,0 +1,121 @@
...
+    "slicer": false
+  },
+  "info": {
+    "use": {
+      "title": "Intended use",
+      "text": "This algorithm analyzes non-contrast CT scans of the thorax and predicts the lung cancer risk. ",
+      "references": [],
+      "tables": []
+    },
+    "analyses": {
```
LennyN95, line 90: Mandatory.

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -0,0 +1,121 @@
...
+      "title": "",
+      "text": "",
+      "references": [],
+      "tables": []
+    },
+    "evaluation": {
```
LennyN95, line 124: Mandatory.

LennyN95 (changes requested):

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -3,7 +3,7 @@
   "name": "lung_cancer_risk_estimation",
   "title": "Lung cancer risk estimation on thorax CT scans",
   "summary": {
-    "description": "This algorithm analyzes non-contrast CT scans of the thorax and predicts the lung cancer risk. ",
+    "description": "This algorithm analyzes non-contrast CT scans of the thorax and predicts the lung cancer risk. The model consists of two modules. The first one is a 3D region proposal network for nodule detection, which outputs all suspicious nodules for a subject. The second one selects the top five nodules based on the detection confidence, evaluates their cancer probabilities and combines them with a leaky noisy-or gate to obtain the probability of lung cancer for the subject",
```
LennyN95: How are the "top five" nodules determined? 
Is this based on model prediction scores? I'm curious whether the algorithm would discard one out of 6 nodules if they all had very similar (but numerically descending) scores and weather such a scenario is realistic or not.

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -71,19 +71,35 @@
...
+          "url": "https://grand-challenge.org/algorithms/dsb2017-grt123/"
+        }
+      ],
       "tables": []
     },
     "analyses": {
-      "title": "",
-      "text": "",
+      "title": "Evaluation",
+      "text": "The nodule detection was evaluated on the validation set of DSB. It contains data from 198 cases and there are 71 (7 nodules smaller than 6 mm are ruled out) nodules in total. The Free Response Operating Characteristic (FROC) is used to evaluate the performance and can be viewed in the publication. The Case classification is evaluated using the Area Under the Curve (AUC) metric.",
```
LennyN95: DSB?

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -71,19 +71,35 @@
...
+          "AUC test set": 0.87
+        }
+      }]
     },
     "evaluation": {
-      "title": "",
-      "text": "",
-      "references": [],
+      "title": "Evaluation data",
+      "text": "The model was evaluated against a private dataset of 300 low-dose CT images. 150 patient scans were from the competition set and 150 were from an independent dataset. Both test datasets contained 50 cancer-positive scans and 100 cancer-negative scans.",
```
LennyN95: If either the competition set or the independent one is publicly available, could we add a reference?
Otherwise, I'm wondering if the distinction is of relevance.

miriam-groeneveld: No, these are private, I'll remove the distinction.

models/gc_grt123_lung_cancer/meta.json
miriam-groeneveld: From the paper: 

>The five proposals are picked out based on their confidence scores in N-Net. As a simple way of data augmentation, during training, proposals are picked stochastically. The probability of being picked for a nodule is proportional to its confidence score.. But during testing, top five proposals are directly picked. If the number of detected proposals is smaller than five, several blank images are used as inputs so that the number is still five.

Not sure how much detail should be added to here, or what is actually meant with "picked directly",

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -0,0 +1,144 @@
...
+        "label": "AUC for the Case classification",
+        "entries": {
+          "AUC training set": "0.90",
+          "AUC test set": "0.87"
+        }
+      }]
+    },
+    "evaluation": {
+      "title": "Evaluation data",
+      "text": "The model was evaluated against a private dataset of 300 low-dose CT images, containing 100 cancer-positive scans and 200 cancer-negative scans.",
```
LennyN95: `300 != 506`, evaluation case numbers missmatching

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -0,0 +1,144 @@
...
+        {
+          "label": "Evaluation paper",
+          "uri": "https://pubmed.ncbi.nlm.nih.gov/34870218/"
+        }
+      ],
+      "tables": []
+    },
+    "training": {
+      "title": "Training data",
+      "text": "The Luna dataset includes the images from the LIDC/IDRI dataset in a different format, together with additional annotations. The LIDC/IDRI dataset is available at https://wiki.cancerimagingarchive.net/display/Public/LIDC-IDRI under a Creative Commons Attribution 3.0 Unported License.",
```
LennyN95: Please use references for all links and write [n] to 'reference' the nth reference.

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -0,0 +1,144 @@
...
+          "label": "LUng Nodule Analysis 2016 dataset part 1",
+          "uri": "https://zenodo.org/record/3723295"
+        },
+        {
+          "label": "LUng Nodule Analysis 2016 dataset part 2",
+          "uri": "https://zenodo.org/record/4121926"
+        },
+        {
+          "label": "Data Science Bowl 2017 dataset",
+          "uri": "https://www.kaggle.com/competitions/data-science-bowl-2017/data"
```
LennyN95: Unclear how this link is relevant. The data has been removed from kaggle.

models/gc_grt123_lung_cancer/meta.json
LennyN95: This description suggests that all data for training comes from https://wiki.cancerimagingarchive.net/display/Public/LIDC-IDRI. 

1. The link is broken but offers a [follow-up](https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=1966254) link. Please us the correct link instead.
2. The website found under the follow-up link reports `1018 cases`. The `LIDC-IDRI_MetaData` csv contains 1308 rows. You report 2285 training cases. It's absolutely not clear to me on what data the model was trained on. Most importantly, the composition and potential selection criteria (if applied) are unclear.

LennyN95 (changes requested):

LennyN95:
Note: test passed (09.01.2024).

```text
DICOM case from NLST on IDC: 
https://portal.imaging.datacommons.cancer.gov/explore/filters/?collection_id=nlst

Case ID: 100002
StudyInstanceUID: 1.2.840.113654.2.55.68425808326883186792123057288612355322
SeriesInstanceUID: 1.2.840.113654.2.55.229650531101716203536241646069123704792

s5cmd --no-sign-request --endpoint-url https://s3.amazonaws.com cp 's3://idc-open-data/b22ed5a6-ad69-4b00-ba26-ae75e96345f8/*' .

Expected output:
---------------
grt123_lung_cancer_findings.json

├── lncancerprob [Lung Nodule cancer probability score.]
│   The predicted cancer probability score for a single lung nodule detected by the algorithm (for nodule 0 at location (112.39999399999999, -92.0, -260.545013))
│   └── Lung Nodule cancer probability score. (0.01483230385929346)
│   ├── id: 0
│   ├── x: 112.39999399999999
│   ├── y: -92.0
│   └── z: -260.545013
├── lncancerprob [Lung Nodule cancer probability score.]
│   The predicted cancer probability score for a single lung nodule detected by the algorithm (for nodule 1 at location (-47.60000600000001, 75.0, -245.54501299999998))
│   └── Lung Nodule cancer probability score. (0.01104212086647749)
│   ├── id: 1
│   ├── x: -47.60000600000001
│   ├── y: 75.0
│   └── z: -245.54501299999998
├── clcancerprob [Case level cancer probability score.]
│   Case level probability score
│   └── Case level cancer probability score. (0.025710642337799072)
│   ├── min: 0.0
│   ├── max: 1.0
│   └── type: probability
```

models/gc_grt123_lung_cancer/meta.json
miriam-groeneveld: You are right, this isn't very clear. The training data consists of:
1. The LuNa dataset. This is a subset of the LIDC-IDRI dataset, excluding scans with a slice thickness greater than 2.5 mm, resulting in a total of 888 CT scans. I can change the link to the dataset of the Luna challenge if that would make things clearer.
2. The Kaggle dataset, which contains `1397` cases in it's training set. This is why there is a link to the dataset, eventhough it's not available anymore. 

> The LIDC-IDRI_MetaData csv contains 1308 rows.

True, but it contains different modalities. When you select only CT, there are `1018` entries, which matches the number of cases reported.

models/gc_grt123_lung_cancer/meta.json
miriam-groeneveld: See comment above

models/gc_grt123_lung_cancer/meta.json
miriam-groeneveld: There are actually 2 evaluations:
1 internal evaluation, with 506 cases.
1 external evaluation (described in the second publication), that uses 300 cases, but 150 of those are from the internal evaluation. How shall we describe that here? They are all private, but the setting for private/public and external apply to all datasets I think?

LennyN95: I suggest the following:
- under `summary.data.evaluation.vol_samples`, report the total number of individual evaluation cases the model was tested on and for which metrics/results are reported on the model card.
- under `info.evaluation` describe in sufficient detail how and on what data the model was tested. Report numbers in a table. For two separate evaluations, you could, e.g. use two tables and explain in the text.

If the internal evaluation is not reported, i.e., no table with statistics, no detailed description, you could instead ignore it altogether and only focus on the public evaluation.

models/gc_grt123_lung_cancer/meta.json
LennyN95: Okay, great, then just update the text with this information. That the dataset is no longer available is no issue per se; please highlight it in the text though, so that we set the expectations right.

models/gc_grt123_lung_cancer/meta.json
silvandeleemput: We have updated the details. We have decided to keep the evaluation samples at 506 which is in line with the reported DSB2017 test set cases and the number of training samples to 2483, which is the LUNA16 dataset (888 cases) + DSB2017 training and validation sets of 1397 and 506 patient cases. 

Results for the external evaluation have been reported separately under evaluation.

models/gc_grt123_lung_cancer/meta.json
silvandeleemput: We updated the information and the links.

silvandeleemput:
We have updated the meta.json, please have a look if you agree. There were no changes to the code, so the test should still pass. We should be able to move on with this.

LennyN95:
There's one last problem showing up here:

When running the model, the following print out is not captured and pollutes the console: `(diag-)image_loader not found, loading dicom will not be possible`, caused by [this line](https://github.com/DIAGNijmegen/bodyct-dsb2017-grt123/blob/9a4ca0415c7fc1d3023a16650bf1cdce86f8bb59/preprocessing/step1.py#L17).

The problem is, that the print statement is executed *at import time*, which is before we (technically can) start capturing prints. 

You may be able to resolve the issue by moving the import inside the `task()` method of the runner `Module`.

For the upcoming models, please always check that the print-out is clean and when running in normal mode (no `--print` or `--debug`), the output is consistent (any uncaptured outputs like this one will break the appearance).

silvandeleemput:
@LennyN95 Good catch. The print statement issue has been addressed by doing as you suggested. I missed the minor glitch when running it in normal mode the first time I checked it. I'll be more aware of the issue with the upcoming models.

LennyN95 (approved):
Requested changes addressed, tests passed.

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -0,0 +1,171 @@
...
+        "slicethickness": "2.5mm",
+        "non-contrast": true,
+        "contrast": false
+      }
+    ],
+    "outputs": [
+      {
+        "type": "Prediction",
+        "valueType": "number",
+        "label": "Cancer probability score",
```
LennyN95: This label will be used as filter attribute on the MHub website and thus must be as specific as possible.

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -0,0 +1,171 @@
...
+        "description": "Probability that the scan contains cancer nodules",
```
LennyN95: It's not the entire scan but just the lungs, right?

LennyN95 (changes requested):
Two more changes required to enhance the integration into the MHub model repository and UX when searching / browsing models on mhub.ai/models.

models/gc_grt123_lung_cancer/meta.json
```diff
@@ -0,0 +1,171 @@
+{
+  "id": "2e67a3cc-4680-4058-bf4e-f965cf50f06f",
+  "name": "lung_cancer_risk_estimation",
```
LennyN95: must match the model name: `gc_grt123_lung_cancer`

LennyN95 (changes requested):

models/gc_grt123_lung_cancer/meta.json
LennyN95: NOTE: I updated our automated checks to prevent a model name mismatch from slipping through.

silvandeleemput: I noticed. That's very convenient.

models/gc_grt123_lung_cancer/meta.json
silvandeleemput: I'll try to make it a bit more specific.

models/gc_grt123_lung_cancer/meta.json
silvandeleemput: Correct, probability that there are any cancer nodules in the lungs. I'll make it more specific.

models/gc_grt123_lung_cancer/meta.json
silvandeleemput: The name has been matched.

LennyN95 (approved):

PR #30, opened 2023-07-18
Update CaSuSt model

LennyN95:
- update casust Dockerfile starting off from new base image
- rename default config from `config.yml` to `default.yml`
- rename generated segmentation files `<model-name/ROI-ID>.seg.dcm`


LennyN95:
tested locally

PR #38, opened 2023-07-25
MHub / GC - Add the GC Tiger challenge winner LB2 model

silvandeleemput:
It took a bit longer than expected, but this PR adds the GC Tiger challenge winner LB2 model to MHub. 
* This implements a Dockerfile based on the new base image
* I added two pipeline files:
    * `default.yml/run.py` (WSI-DICOM in JSON out) <--- This one uses the PanImgConverter
    * `tiff_pipeline.yml/run_tiff_pipeline.py` (TIFF in JSON out) <-- This one was added for convenience/comparison/testing
* The PR should probably be retargeted to an `m-gc-tiger-lb2` branch, subsequently, the MHub model repo should be properly added to the Dockerfile

This is also an interesting model to have a look at because of the following issues/characteristics:
  * Uses WSI as an input and uses the PanImgConverter
  * Has a JSON output with a single integer value, which is the TIL score. 
  * The TigerLB2Runner requires a GPU to be run (enforced through an assert, otherwise it would throw an error). It will require elaborate hacking to get working on a CPU (I have been able to do it for a CPU-only version), but the runtime is over one hour for the CPU version versus 15-20 mins on the GPU so I think not worth supporting it. Let me know otherwise.
  * ~~This model suffers from the model import issue, (check the Dockerfile), which required me to perform some ugly hacks that even prevent me from running the pipeline directly from the configuration files.~~


silvandeleemput:
Pushed a fix for the model import issue as suggested by @LennyN95 to this PR. 
To circumvent the conflicting imports issue I added a CLI script for running the model algorithm that is now called as a subprocess within the pipeline. I consider the issue solved for this particular model.

LennyN95:
Can we use ValueOutput+ReportExporter for this algorithm? 
If the JSON generation is characteristic to the algorithm, I suggest keeping the JSON file as a data output and adding a ValueOutput with the score for greater flexibility and compatibility. This way, we allow users to either use the original file and simultaneously offer full flexibility via our ReportExporter mechanisms.

silvandeleemput:
This PR has been cleaned and updated to meet the latest requirements:
- Dockerfile
  - fixed to commit for tiger_challenge repo
  - updated MHubAI/models clone stub
  - updated entrypoint
- pipeline default.yml
  - updated the outputs
  - use the TiffConverter
  - added reportexporter settings
- runner:
  - Added TIL Score as ValueOutput
  - Confirmed data output
  - Added/updated comments
- removed:
  - custom panimg converter classes
  - run.py script to run the pipeline

> Can we use ValueOutput+ReportExporter for this algorithm? If the JSON generation is characteristic to the algorithm, I suggest keeping the JSON file as a data output and adding a ValueOutput with the score for greater flexibility and compatibility. This way, we allow users to either use the original file and simultaneously offer full flexibility via our ReportExporter mechanisms.

@LennyN95 I added a ValueOutput for the TIL score for this algorithm, but when I try to export that value with the ReportExporter it outputs an empty report with: `{}` I get the following error using the `--print` option: 

```
Start ReportExporter
Error while generating report for  {'data': 'til_score', 'label': 'TIL score', 'value': 'value'}
```

After doing some in-situ debugging, I found that the following line fails:

`mhubio/modules/exporter/ReportExporter.py::153:  assert len(data) == 1, "data name not unique."` 

Then I found out that `data = []` is an empty list after printing the data, which was populated using the following code:
```
# fetch the data
data = [d for d in instance.outputData if d.name == include['data']]
```

It appears that `instance.outputData = []` is also empty, so it seems that the outputData is somehow not properly picked up when using this decorator in the TigerLB2Runner class:

`@IO.OutputData('til_score', TilScoreOutput, data='in_data', the='TIGER LB2 TIL score - percentage of stromal area covered by tumour infiltrating lymphocytes. Values between 0-100 (percent).')`


@LennyN95 Do you know what is going on here? The latest code update in this PR is the code that I used. I only disabled the ReportExporter in the default.yml for now to make this fully functional, but during testing this was enabled. 

I got it working with similar code before for other model PRs. It doesn't however seem to be caused by codebase changes, because I tried to revert it to older versions of mhubio up until August 3th, but it still had the same output and error. So I guess it is something else.

silvandeleemput:
Alright, I got it working, this PR can now use the ReportExporter as requested. 

Apparently it was a mistake in the last `assert` line of the runner (called the wrong method). The error message was reported in the output using `--print`, but because the code doesn't stop at the first error, the rest of the output obfuscated this error message. I think this further emphasizes the need for a `--stop-on-error` flag to avoid unnecessary debugging.

models/gc_tiger_lb2/utils/TigerLB2Runner.py
```diff
@@ -0,0 +1,59 @@
...
+import numpy as np
+import SimpleITK as sitk
+import torch
+
+import sys
+import json
+
+
+@ValueOutput.Name('til_score')
+@ValueOutput.Meta(Meta(key="value"))
```
LennyN95: This is a placeholder.
Consider adding valuable meta-data such as the value format (e.g., scale=percent, min=0, max=100)

models/gc_tiger_lb2/utils/TigerLB2Runner.py
```diff
@@ -0,0 +1,59 @@
...
+        self.subprocess(
+            [
+                sys.executable,
+                str(self.CLI_SCRIPT_PATH),
+                in_data.abspath,
+                out_data.abspath,
+            ]
+        )
+
+        if not Path(out_data.abspath).is_file():
```
LennyN95: This is not required; our `IO.Output` decorator performs a file-existence check and confirms only existing files using the official data API (`data.confirm()`).

LennyN95: If required for a subsequent step, pls invert the expression instead of throwing exceptions/errors.

models/gc_tiger_lb2/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    rm -rf /vuno/.git
+
+# Install tiger LB2 dependencies
+RUN pip3 install --no-cache-dir -r /vuno/requirements.txt
+
+# Reinstall correct version of Numpy to function with ASAP 2.1
+RUN pip3 install --no-cache-dir --force-reinstall numpy==1.22
+
+# Enforce minimum version of panimg (with WSI fix)
+RUN pip3 install --no-cache-dir panimg>=0.13.2
```
LennyN95: Did you encounter any issues without specifying this line? 
The panimg from our base image should already satisfy the requirement.

models/gc_tiger_lb2/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+#   - Subsequently we remove the .git directory to procuce a compacter docker layer
+RUN git clone https://github.com/vuno/tiger_challenge.git /vuno && \
+    cd /vuno && git reset --hard 720f8dfca4624792c8e57915c4222efec5a0c2d4 && \
+    rm -rf /vuno/.git
+
+# Install tiger LB2 dependencies
+RUN pip3 install --no-cache-dir -r /vuno/requirements.txt
+
+# Reinstall correct version of Numpy to function with ASAP 2.1
+RUN pip3 install --no-cache-dir --force-reinstall numpy==1.22
```
LennyN95: This might clash with future mhubio updates. One option as once discussed could be wrapping up all algorithm dependencies in a pip env and then calling the cli script in that context.

models/gc_tiger_lb2/config/default.yml
```diff
@@ -0,0 +1,31 @@
...
+- ReportExporter
+- DataOrganizer
+
+modules:
+  DicomImporter:
+    source_dir: input_data
+    import_dir: sorted_data
+    sort_data: True
+    meta: 
+      mod: sm
```
LennyN95: `mod: "%Modality"` can be used to automatically parse the dicom Modality header.

models/gc_tiger_lb2/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+# Install ASAP
+RUN apt-get update \
+    && apt-get -y install curl libpython3.8-dev \
+    && curl --remote-name --location "https://github.com/computationalpathologygroup/ASAP/releases/download/ASAP-2.1/ASAP-2.1-py38-Ubuntu2004.deb" \
+    && dpkg --install ASAP-2.1-py38-Ubuntu2004.deb || true \
+    && apt-get -f install --fix-missing --fix-broken --assume-yes \
+    && ldconfig -v \
+    && apt-get clean \
+    && echo "/opt/ASAP/bin" > /usr/local/lib/python3.8/dist-packages/asap.pth \
```
LennyN95: Looks a bit hacky. Is this the official / stable install?

models/gc_tiger_lb2/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# Enforce minimum version of panimg (with WSI fix)
+RUN pip3 install --no-cache-dir panimg>=0.13.2
+
+# Download and install model weights file from zenodo
+RUN rm -rf /vuno/pretrained_weights && \
+    wget https://zenodo.org/record/8112176/files/pretrained_weights.zip -O /vuno/pretrained_weights.zip && \
+    unzip /vuno/pretrained_weights.zip -d /vuno && \
+    rm /vuno/pretrained_weights.zip
+
+# Clone the main branch of MHubAI/models TODO
```
LennyN95: To avoid disabling the model import during development, please use the recently introduced model import build utility instead:

```
ARG MHUB_MODELS_REPO
RUN buildutils/import_mhub_model.sh gc_tiger_lb2 ${MHUB_MODELS_REPO}
```

LennyN95 (changes requested):

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: Done!

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: Yes, there was a fix in that specific version, and wasn't part of the base image at the time, it is redundant now and I removed it.

models/gc_tiger_lb2/utils/TigerLB2Runner.py
silvandeleemput: Ok, modified.

models/gc_tiger_lb2/utils/TigerLB2Runner.py
silvandeleemput: I think doing it this way is preferable since the raised error clearly communicates what the issue is within the module. If the expression is inverted it would become a silent failure, which is much harder to debug.

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: Could be, but the other modules besides the runner used for this model don't use Numpy as far as I can tell correct? I rather not touch this, since this will be very hard to configure correctly in combination with ASAP.

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: Yes, this code came with the Dockerfile used for the Tiger challenge.

models/gc_tiger_lb2/config/default.yml
silvandeleemput: I rather keep it at `sm`, since that should be the correct DICOM modality anyway for this pipeline to run as all inputs should be slide microscopy images. Is there any benefit to changing it here, except that nothing runs if it does not match the modality?

LennyN95: Yes, e.g. Patient folder contains two dicom series, one SM and the other CT. It makes no sense to import both as SM. 

Please change as suggested.

models/gc_tiger_lb2/dockerfiles/Dockerfile
LennyN95: Will you go back and fix it the moment we use Numpy in e.g. calculate general run-time statistics :)?
I already made a suggestion, you can install ASAP and the fixed Numpy version in a pip-env.

Please update before we can proceed.

models/gc_tiger_lb2/utils/TigerLB2Runner.py
LennyN95: `Something went wrong` does not at all `clearly communicate what the issue` is.  
Either improve or update as suggested.

models/gc_tiger_lb2/dockerfiles/Dockerfile
LennyN95: Note, that if the Ubuntu or the Python version changes, this model will become deprecated or require a revision.

Do you know if ASAP can run on Ubuntu 22.04 or with any other Python version?
Did you think about a maintenance strategy?

models/gc_tiger_lb2/meta.json
```diff
@@ -0,0 +1,113 @@
...
+        {
+          "label": "On the C-statistics for Evaluating Overall Adequacy of Risk Prediction Procedures with Censored Survival Data",
+          "uri": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3079915/"
+        }
+      ],
+      "tables": []
+    },
+    "evaluation": {
+      "title": "Evaluation data",
+      "text": "The test set consists of a separate dataset of n=707 H&E-stained whole-slide breast cancer histopathology images",
```
LennyN95: Is the mentioned dataset closed-source? Please specify or provide a public reference.#
Why is evaluation `summary.data.evaluation.vol_samples = 58 << n = 707`?

models/gc_tiger_lb2/meta.json
```diff
@@ -0,0 +1,113 @@
+{
+  "id": "c5397909-0397-489f-8744-6bf3952e9a1c",
+  "name": "tiger_lb2",
+  "title": "TIGER challenge winner: Team VUNO",
+  "summary": {
+    "description": "Participants in the TIGER challenge will have to develop computer algorithms to analyze H&E-stained whole-slide images of breast cancer histopathology, to perform three tasks:  detection of lymphocytes and plasma cells, which are the main types of cells considered as tumor-infiltrating lymphocytes; segmentation of invasive tumor and tumor-associated stroma, which are the main tissue compartments considered when identifying relevant regions for the TILs; compute an automated TILs score, one score per slide, based on the output of detection and segmentation.",
```
LennyN95: This is misleading and reads like this is an ongoing challenge. It is not clear that this is about a particular (winning) model from a past challenge.

LennyN95 (changes requested):

models/gc_tiger_lb2/meta.json
silvandeleemput: We will update this.

models/gc_tiger_lb2/meta.json
silvandeleemput: We will update this.

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: Hi, yes as discussed: There is a newer version of ASAP for Ubuntu 22.04, so once the base image moves, we can move along. The maintenance strategy is to support these Ubuntu and Python versions (potentially in a pyenv), once 22.04 is rolled out we can update accordingly. For steps beyond that we have to discuss...

models/gc_tiger_lb2/config/default.yml
silvandeleemput: Ok, I'll update it as suggested.

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: I'll try to install them in a pip-env, it'll give some overhead, but less maintainability issues.

models/gc_tiger_lb2/utils/TigerLB2Runner.py
silvandeleemput: I'll change the error message to a better one.

models/gc_tiger_lb2/dockerfiles/Dockerfile
```diff
@@ -20,9 +20,11 @@ RUN apt-get update \
 # Install tiger LB2 algorithm
 #   - Clone tiger LB2 codebase (master branch, fixed to commit 720f8dfca4624792c8e57915c4222efec5a0c2d4)
 #   - Subsequently we remove the .git directory to procuce a compacter docker layer
+#   - We also fix a reproducibility issue with the segmentation prediction by fixing the torch random seed
 RUN git clone https://github.com/vuno/tiger_challenge.git /vuno && \
     cd /vuno && git reset --hard 720f8dfca4624792c8e57915c4222efec5a0c2d4 && \
-    rm -rf /vuno/.git
+    rm -rf /vuno/.git && \
+    sed -i "238i \        torch.manual_seed(0)" /vuno/segmentation/inference.py
```
LennyN95: This is great but should be implemented in the source repository.

LennyN95 (changes requested):

models/gc_tiger_lb2/meta.json
silvandeleemput: This was changed to the correct value.

models/gc_tiger_lb2/meta.json
silvandeleemput: We have rewritten it to be more like a description of the algorithm.

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: Ok, it is now part of a pipenv.

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: It has been installed inside a pipenv.

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: Yes, ideally we would be able to make the necessary changes in the original repository. However, since it is unlikely that we have that option, we need to discuss the best approach to take. One possible solution is forking the repository and implementing the changes in the forked version. Let's discuss this in more detail during our next meeting. For now, let's focus on the remaining changes and come up with a solid plan to address this issue as soon as possible.

silvandeleemput:
[output.zip](https://github.com/MHubAI/models/files/14526578/output.zip)

silvandeleemput:
/test

```yaml
sample:
  idc_version: 17.0
  data:
    - SeriesInstanceUID: 1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0
      aws_url: s3://idc-open-data/20ab74ef-080b-4975-adad-7507b7b04cb8/*
      path: dicom

reference:
  url: https://github.com/MHubAI/models/files/14526578/output.zip
```

---

<details><summary>Test Results (24.03.13_17.52.53_6VEUMKpYVb)</summary>

```yaml
id: 7654bec5-0c56-43b5-82db-cee986e4d663
date: '2024-03-13 18:08:50'
checked_files:
- file: gc_tiger_lb2_til_score.json
  path: /app/test/src/1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0/gc_tiger_lb2_til_score.json
  checks:
  - checker: DataFileCheck
    findings:
    - label: Value similar
      description: Value of key 'TIL score' is similar in source file
      subpath: TIL score
      info:
        scale: 1
        precision: -1
        src: 4
        ref: 5
  - checker: SizeCheck
summary:
  files_missing: 0
  files_extra: 0
  checks:
    DataFileCheck:
      files: 1
      findings:
        Value similar: 1
    SizeCheck:
      files: 1
conclusion: false
```

</details>

silvandeleemput:
/review

We have implemented the requested changes and updated the meta.json. 
Except for the reproducibility fix it should be as good as ready.

silvandeleemput:
@LennyN95 thanks for checking. Small differences in outcome are to be expected when run on different GPUs due to accumulated small numerical differences. If the results only differ by a few points (something in the range of 1-2 points) this is deemed acceptable for this model as discussed with the challenge organizers.

models/gc_tiger_lb2/dockerfiles/Dockerfile
silvandeleemput: I have forked the repository and added the change there. The manual correction is no longer in the Dockerfile.

silvandeleemput:
[gc_tiger_lb2_output.zip](https://github.com/MHubAI/models/files/15091074/gc_tiger_lb2_output.zip)

silvandeleemput:
@LennyN95 I have created a fork of the source model/algorithm repository here: https://github.com/DIAGNijmegen/tiger_vuno_algorithm. 

In this fork, I have added the reproducibility fix and added the tumor and stroma segmentation mask to the output.
The PR has been updated to build from the fork and output the segmentation.

I have uploaded a new output.zip file (which includes the segmentation) and updated the URL in the test above.

Please check if everything is in order now. I suspect that the segmentation will be different due to GPU hardware/driver differences.

LennyN95:
Thank you @silvandeleemput. I create this as a new test and reverted the modification to keep the test history clean. Test requests should be edited to correct syntax errors only but not touched once one or more tests were performed (and reported in the test request).

LennyN95:
/test

```yaml
sample:
  idc_version: 17.0
  data:
    - SeriesInstanceUID: 1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0
      aws_url: s3://idc-open-data/20ab74ef-080b-4975-adad-7507b7b04cb8/*
      path: dicom

reference:
  url: https://github.com/MHubAI/models/files/15091074/gc_tiger_lb2_output.zip
```

---

<details><summary>Test Results (24.04.25_08.24.29_0ZZ5RYzDlk)</summary>

```yaml
id: fa0fa712-4d2b-4f13-adae-56cc312c4209
date: '2024-04-25 08:42:26'
missing_files:
- 1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0/gc_tiger_lb2_segmentation.mha
checked_files:
- file: gc_tiger_lb2_til_score.json
  path: /app/test/src/1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0/gc_tiger_lb2_til_score.json
  checks:
  - checker: DataFileCheck
    findings:
    - label: Value similar
      description: Value of key 'TIL score' is similar in source file
      subpath: TIL score
      info:
        scale: 1
        precision: -1
        src: 4
        ref: 5
  - checker: SizeCheck
summary:
  files_missing: 1
  files_extra: 0
  checks:
    DataFileCheck:
      files: 1
      findings:
        Value similar: 1
    SizeCheck:
      files: 1
conclusion: false
```

</details>

---

<details><summary>Test Results (24.04.25_14.08.51_xHF1bsAg8N)</summary>

```yaml
id: 1c55ef8c-8325-4838-898e-63a7ef2b04fe
date: '2024-04-25 14:32:07'
checked_files:
- file: gc_tiger_lb2_segmentation.mha
  path: /app/test/src/1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0/gc_tiger_lb2_segmentation.mha
  checks:
  - checker: ImageFileCheck
    notes:
    - label: Data Type
      info: uint8
    findings:
    - label: Exception
      description: An exception occurred during check
      info: 'could not convert string to float: '''''
- file: gc_tiger_lb2_til_score.json
  path: /app/test/src/1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0/gc_tiger_lb2_til_score.json
  checks:
  - checker: DataFileCheck
    findings:
    - label: Missing key
      description: Key 'TIL score' is missing in source file
      subpath: TIL score
  - checker: SizeCheck
    findings:
    - label: Size Difference
      description: file size is smaller than reference
      info:
        src_size: 2
        ref_size: 22
        diff_size: -20
summary:
  files_missing: 0
  files_extra: 0
  checks:
    ImageFileCheck:
      files: 1
      findings:
        Exception: 1
    DataFileCheck:
      files: 1
      findings:
        Missing key: 1
    SizeCheck:
      files: 1
      findings:
        Size Difference: 1
conclusion: false
```

</details>

---

<details><summary>Test Results (24.06.03_10.26.45_9t3n5BDtdf)</summary>

```yaml
id: bc76f65a-b934-4712-a8cf-2c5263f357ae
date: '2024-06-03 10:54:38'
checked_files:
- file: gc_tiger_lb2_segmentation.mha
  path: /app/test/src/1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0/gc_tiger_lb2_segmentation.mha
  checks:
  - checker: ImageFileCheck
    notes:
    - label: Data Type
      info: uint8
    findings:
    - label: Exception
      description: An exception occurred during check
      info: 'could not convert string to float: '''''
- file: gc_tiger_lb2_til_score.json
  path: /app/test/src/1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0/gc_tiger_lb2_til_score.json
  checks:
  - checker: DataFileCheck
    findings:
    - label: Value similar
      description: Value of key 'TIL score' is similar in source file
      subpath: TIL score
      info:
        scale: 1
        precision: -1
        src: 4
        ref: 5
  - checker: SizeCheck
summary:
  files_missing: 0
  files_extra: 0
  checks:
    ImageFileCheck:
      files: 1
      findings:
        Exception: 1
    DataFileCheck:
      files: 1
      findings:
        Value similar: 1
    SizeCheck:
      files: 1
conclusion: false
```

</details>

---

<details><summary>Test Results (24.06.07_12.29.30_uTzBQF9Njk)</summary>

```yaml
id: 3302c113-2efa-45e3-97db-7e651d4f835f
date: '2024-06-07 12:46:36'
checked_files:
- file: gc_tiger_lb2_segmentation.mha
  path: /app/test/src/1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0/gc_tiger_lb2_segmentation.mha
  checks:
  - checker: ImageFileCheck
    notes:
    - label: Data Type
      info: uint8
    - label: Dice Score
      info: 1.0
- file: gc_tiger_lb2_til_score.json
  path: /app/test/src/1.3.6.1.4.1.5962.99.1.1334438926.1589741711.1637717011470.2.0/gc_tiger_lb2_til_score.json
  checks:
  - checker: DataFileCheck
    findings:
    - label: Value similar
      description: Value of key 'TIL score' is similar in source file
      subpath: TIL score
      info:
        scale: 1
        precision: -1
        src: 4
        ref: 5
  - checker: SizeCheck
summary:
  files_missing: 0
  files_extra: 0
  checks:
    ImageFileCheck:
      files: 1
    DataFileCheck:
      files: 1
      findings:
        Value similar: 1
    SizeCheck:
      files: 1
conclusion: false
```

</details>

silvandeleemput:
@LennyN95 Regarding the results, the exact same exception as last time shows:
```    notes:
    - label: Data Type
      info: uint8
    findings:
    - label: Exception
      description: An exception occurred during check
      info: 'could not convert string to float: '''''
```

I don't understand the origin of this error yet and it wasn't present two base images ago. I will investigate this further.

silvandeleemput:
@LennyN95 The results look fine now.

LennyN95 (approved):

PR #43, opened 2023-08-21
BAMF - NNUnet Liver CT

rahul99:


LennyN95:
Hey Rahul, thx for the PR! I'm currently on vacation, so I cannot directly test your code but looks good (and we've been through most anyways together).

Here are some comments
- [ ] Scripts can be removed 
- [x] The 'source_segs' attribute for the DsegConverter can be changed to a string (arrays are still supported but will raise a deprecated warning) 
- [x] Are the weights on Google version controlled? It's required to pull the weights from a fixed revision such that they cannot be changed unless the Dockerfile is modified to reflect that. We suggest uploading the weights in Zenodo (but as long as our requirement is fulfilled, any other service is just fine). 
- [x] Model name set for DsegConverter and used in file names in the DataOrganizer should be unique so that all MHub models can be distinguished
- [x] The entry point (CMD) in the Dockerfile points to the wrong path

rahul99:
Hey Leo, I have made the changes as per your comments. Some details below:

- We have a post-processing module for all segments. So we are adopting scripts as an entrypoint for container
- Changed the `source_segs` from list to string
- Moved weights to Zenodo
- Unique model names set for Dseg converter and organizer
- entry point fixed

LennyN95:
Hi Rahul,

Great! 

I've seen your processor but that's no problem, you can still use the config entry point. Please create a file like "BamfProcessor.py" in utils, then define a single BamfProcessor(Module) class in that file so it will be available in our execute chain defined via the config files. You can have a look on the most recent runner modules, e.g. those of the GC LungLobes or the CaSuSt model. 

The meta json should be removed as the ROI can be defined with a SegDB id (LIVER) and we can auto generate the json. This is always preferred over a custom json for Dicomseg as we need full control over generated output to guarantee standardization. The option to provide a custom json is only for testing until we merge those structures into SegDB. 

Great work so far, I'm back from vacation on Monday and will give them a test run. If helpful, we can schedule a meeting from Wednesday ;) 

Best,
Leo.

rahul99:
Hey Leo, closing this one as I have created a new PR where branch name is also aligned with the task

PR #46, opened 2023-08-29
MHub / GC - Add LUNA22 ISMI baseline model implementation

silvandeleemput:
This PR adds the GC LUNA22 ISMI baseline model (from the [LUNA22 challenge](https://luna22-ismi.grand-challenge.org/)) to MHub.
GitHub Repo: https://github.com/DIAGNijmegen/bodyct-luna22-ismi-algorithm-baseline
GC page: https://grand-challenge.org/algorithms/luna22-ismi-baseline/

### Algorithm I/O
  * Input should be a volume from a CT Lung image (MHA) centered around a single nodule (This MHub implementation expects a Dicom which is converted to MHA using MhaConverter)
  * Outputs are:
    * A JSON file with the following output scores:
      * malignancy risk
      * lung nodule type
  * Test images for this algorithm can be downloaded from here: https://zenodo.org/record/6559584

### Caveats
* The PR target is `main`, but should be something like `m-gc-luna22-ismi-baseline`
* ~Added a TODO to the `MHAConverter` since it doesn't have the `panimg` backend yet.~


silvandeleemput:
This algorithm/model should not be added since it is not very suitable as a MHub model, due to it being intended for students. Also, the algorithm focuses on small volumes around nodules, requiring additional preprocessing. Hence we will replace this algorithm with #60, which is a better candidate.

PR #48, opened 2023-08-30
MHub / GC - Add ODE Colon Glands Segmentation Model

silvandeleemput:
This PR adds the ODE Colon Glands Segmentation Model to MHub.
GitHub Repo: https://github.com/DIAGNijmegen/neural-odes-segmentation/tree/master

### Algorithm I/O
  * Input should be a small WSI slide color bitmap similar to the images from the Warwick QU Dataset (MHA), MHA input can either be a 2D volume with RGB color, or a 3D volume with (x,y,c) The Dicom input format should match the 3D volume input format.
  * Output:
    * A MHA Segmentation of the Colon Glands within the input image, every Colon Gland receives a separate ID
  * Train/Test images (Warwick QU Dataset) for this algorithm can be downloaded from here: https://warwick.ac.uk/fac/sci/dcs/research/tia/glascontest
  
### Caveats
* The PR target is `main`, but should be something like `m-gc-ode-colon-glands`
* No Dicom Seg output is given because the Colon Glands require separate IDs per segmentation, but are all of the same class, which is supported by DicomSeg, but no ready-to-use conversion tool exists for this.


fedorov:
> No Dicom Seg output is given because the Colon Glands require separate IDs per segmentation, but are all of the same class, which isn't well supported by DicomSeg yet.

It is perfectly valid to have multiple segments of the same kind in DICOM SEG.

The issue is a bit different - there is no ready-to-use converter for creating DICOM SEG for SM. Those need to be treated differently because of multiresolution, and different conversion process compared to radiology. Let me discuss this with members of the team, and I will confirm what is our current consensus re guidance on this.

fedorov:
Oh, I see - you are referring to a collection that does not exist in IDC, and is restricted access in TCIA.

Again, every algorithm in the GC package should have been selected with the requirement in mind that there should be applicable images in IDC that would work with that algorithm. All images in IDC are in DICOM. So either someone did not confirm that the requirement is met, or someone selected this collection by mistake. In either case - we should not be relying on this collection.

LennyN95:
_**NOTE**: implementation is on hold until preprocessing is clarified._

PR #52, opened 2023-09-05
MHub / GC - Add AutoPET challenge winner model/algorithm

silvandeleemput:
This PR adds the AutoPET challenge winner model (from the [AutoPET  challenge](https://autopet.grand-challenge.org/)) to MHub.
GitHub Repo: https://github.com/YigePeng/AutoPET_False_Positive_Reduction
GC page: https://grand-challenge.org/algorithms/autopet_submit/

### Algorithm I/O
  * Input:
    * A whole-body CT image (MHA)
    * A whole-body PET image (MHA) registered to CT
  * Outputs are:
    * A binary segmentation mask of the malignant lesions in the whole-body CT/PET inputs
    * TCIA dataset (Dicom): https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=93258287
    * (Mha, small test set) https://github.com/YigePeng/AutoPET_False_Positive_Reduction/tree/main/test

### Caveats
* ~The PR target is `main`, but should be something like `m-gc-autopet-fpr`~
* ~In the Dockerfile the MHub model repo integration is currently marked `TODO` since it requires the creation of the appropriate branch for this code first.~
* ~Support for Dicom should be easy to add, but this requires additional data clearance from TCIA, so we haven't tested this yet.~


fedorov:
> Support for Dicom should be easy to add, but this requires additional data clearance from TCIA, so we haven't tested this yet.

I do not understand this. What clearance from TCIA is required, and how is it related to DICOM support?

fedorov:
Oh, I see - you are referring to a collection that does not exist in IDC, and is restricted access in TCIA.

https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=93258287#932582870c7caa21e8b840a393398eeda1279f3b

image

Again, every algorithm in the GC package should have been selected with the requirement in mind that there should be applicable images in IDC that would work with that algorithm. All images in IDC are in DICOM. So either someone did not confirm that the requirement is met, or someone selected this collection by mistake. In either case - we should not be relying on this collection.

silvandeleemput:
Hi, I tried to add a full Dicom pipeline for this model. The input works fine but I lack some information for generating the output Dicom segmentation. It seems that it should be relatively simple to add this since it is a binary segmentation mask of malignant lesions in whole-body CT/PET images. However, I was not able to find something like a general "malignant lesions/cancer" output segmentation class type in the [internal database of segments](https://github.com/MHubAI/SegDB) and also not using the [qiicr webinterface](http://qiicr.org/dcmqi/#/seg). @fedorov Could you help me with this, what segmentation class would you recommend? Or should such a class be created?

fedorov:
@silvandeleemput you just need to select category "morphologically altered structure" in the web app, as below. You can use "neoplasm" as segmentation type, I believe.

image

models/gc_autopet_fpr/config/default.yml
```diff
@@ -0,0 +1,32 @@
...
+- TciaDicomProcessor
+- AutoPETRunner
+- DsegConverter
+- DataOrganizer
+
+modules:
+  FileStructureImporter:
+    input_dir: 'input_data'
+    structures:
+      - $sid@instance/re:.*::/re:.*-GK.*::@dicom:mod=ct
```
LennyN95: I guess this structure comes from the test data, however, we should apply some standards to the default input structure for multi-input modes.

I suggest `$instance/$group`, where `$group` is whatever distinguishes the model inputs (so GK and PET in this case).

For example, the default data then must be organized like so:

```
patient1/
├── GK/
│   ├── 1.dcm
│   ├── 2.dcm
│   └── ...
├── PET/
│   ├── 1.dcm
│   ├── 2.dcm
│   └── ...
patient2/
├── GK/
└── PET/
```

And we can use a less specific and easier to understand importer structure like so:

```yml
structures:
- $sid@instance/GK@dicom:mod=ct
- $sid/PET@dicom:mod=pet
```

---

Alternatively, a dynamic import can be used on the described file structure (the following example resolves to `dicom:mod=ct:type=GK` and `dicom:mod=ct:type=PET`).

```yml
- $sid@instance/$type@dicom:mod=ct
```

This won't fit the described use case as we've two different modalities but you could consider using CT and PET as `$group` in your file structure and replace `$type` by `$mod` in above example.

models/gc_autopet_fpr/utils/TciaDicomProcessor.py
```diff
@@ -0,0 +1,38 @@
...
+    @IO.Output('out_data_pet', 'PET.nii.gz', 'nifti:mod=pt', bundle='tcia', the='Converted PET image')
+    @IO.Output('out_data_ct', 'CTres.nii.gz', 'nifti:mod=ct', bundle='tcia', the='Converted CT image resampled to PET image resolution and size')
+    def task(self, instance: Instance, in_data_ct: InstanceData, in_data_pet: InstanceData, out_data_pet: InstanceData, out_data_ct: InstanceData) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            tmp = Path(tmp)
+            # Convert PET and CT data to Nifti
+            dcm2nii_CT(Path(in_data_ct.abspath), tmp)
+            dcm2nii_PET(Path(in_data_pet.abspath), tmp)
+            # Resample/register CT to PET
+            resample_ct(tmp)
```
LennyN95: Couldn't we use plastimatch for this?
We already have plastimatch and it's python wrapper pyplastimatch installed in our base image.

```python
import pyplastimatch as pypla

resample_args = {
   'input': in_data_ct.abspath,
   'output': out_data_ct.abspath,
   'fixed': in_data_pet.abspath,
}

pypla.resample(
   verbose=self.config.verbose,   
   path_to_log_file=log_data.abspath,
   **resample_args
)
```

models/gc_autopet_fpr/dockerfiles/Dockerfile
```diff
@@ -34,6 +34,16 @@ RUN mkdir -p /opt/algorithm/nnUNet_raw_data_base/nnUNet_raw_data/Task504_Total_P
 ENV nnUNet_raw_data_base="/opt/algorithm/nnUNet_raw_data_base"
 ENV RESULTS_FOLDER="/opt/algorithm/checkpoints"
 
+# install DICOM TCIA preprocessing (registration and nifti conversion)
+RUN git clone https://github.com/lab-midas/TCIA_processing /tmp/algorithm && \
```
fedorov: There is a robust converter from DICOM to NIfTI, which we are already using: dcm2niix. I do not see any reason to experiment with any other tools. I strongly recommend this is taken out and replaced with dcm2niix.

silvandeleemput: This particular processing performs the necessary registration of the CT to the PET scan as well and comes with the data provided on TCIA, I recommend you use this instead of the dcm2nixx tool.

LennyN95: @silvandeleemput @fedorov I addressed this point, too (although in the code implementation). I suggest we try to substitute tcia_processing with plastimatch, which (similar to dcm2niix) is already installed in our base images and can perform the conversion and the registration).

fedorov: Can you point me to the details about that registration? I do not see any mention of "registration" in either README file, or in the code in the repo.

That repo was literally developed for that one specific collection. It is unacceptable for general purpose model to use collection-specific code. The whole point is to make the model robust to any similar dataset. 

I strongly disagree with the suggestion to use that tool. Instead, I recommend you handle CT and PET as two separate channels, similar to how MR T2/ADC are handled in #60.

silvandeleemput: It is true that it isn't mentioned very well in the README, but it's there:
* https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=93258287: at the bottom under "Additional Resources for this Dataset" mentions: "Scripts provided by the submitting group for file conversion, preprocessing alignment and resampling of PET, CT and mask data"
* In the repo https://github.com/lab-midas/TCIA_processing:
  * In the README.md the output for the "TCIA_processing: tcia_nifti_to_mha.py" mentions a CTres.nii.gz (CT image resampled to PET image resolution and size) (see the output image)
  * The code for this can be found here: https://github.com/lab-midas/TCIA_processing/blob/fb7b0740bcab896ed8c8e646c4491cadafbb8789/tcia_dicom_to_nifti.py#L139-L144

I am also convinced that for this dataset the preprocessing pipeline should work for other similar data if these were available (currently no other IDC dataset is a good match). The conversion and registration code will likely work if similar data is presented.

fedorov: @silvandeleemput the code you referenced does not do any registration - it is just the resampling. 

> I am also convinced that for this dataset the preprocessing pipeline should work for other similar data if these were available (currently no other IDC dataset is a good match). The conversion and registration code will likely work if similar data is presented.

I love your optimism, but (just after a very quick look around) I found this line that seems to indicate that the world is not perfect: https://github.com/lab-midas/TCIA_processing/blob/fb7b0740bcab896ed8c8e646c4491cadafbb8789/tcia_dicom_to_nifti.py#L124. And the things they do after that line are just a recipe for a disaster: https://github.com/lab-midas/TCIA_processing/blob/fb7b0740bcab896ed8c8e646c4491cadafbb8789/tcia_dicom_to_nifti.py#L127-L128

> (currently no other IDC dataset is a good match)

Just to set the record straight, there are a lot of collections in IDC that contain PET/CT studies: https://portal.imaging.datacommons.cancer.gov/explore/filters/?Modality_op=AND&Modality=PT&Modality=CT

silvandeleemput: > @silvandeleemput the code you referenced does not do any registration - it is just the resampling.
> 

My bad. Where I said registration should have been resampling, that's correct.

> > I am also convinced that for this dataset the preprocessing pipeline should work for other similar data if these were available (currently no other IDC dataset is a good match). The conversion and registration code will likely work if similar data is presented.
> 
> I love your optimism, but (just after a very quick look around) I found this line that seems to indicate that the world is not perfect: https://github.com/lab-midas/TCIA_processing/blob/fb7b0740bcab896ed8c8e646c4491cadafbb8789/tcia_dicom_to_nifti.py#L124. And the things they do after that line are just a recipe for a disaster: https://github.com/lab-midas/TCIA_processing/blob/fb7b0740bcab896ed8c8e646c4491cadafbb8789/tcia_dicom_to_nifti.py#L127-L128
> 

This is indeed very bad for generalization to other datasets, good catch! It is of course possible to use the dcm2niix tool first and subsequently only use the resampling part of the TCIA codebase (might not even require the TCIA preprocessing install as it seems to be mostly nifti calls). Only the resampling part is essential to the model. We should probably test if both tools produce the same image data on the TCIA dataset though.

> > (currently no other IDC dataset is a good match)
> 
> Just to set the record straight, there are a lot of collections in IDC that contain PET/CT studies: https://portal.imaging.datacommons.cancer.gov/explore/filters/?Modality_op=AND&Modality=PT&Modality=CT

I agree, but the model was trained and validated on the [(FDG-PET-CT-Lesions)](https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=93258287) dataset specifically, the model might not generalize that well to other datasets. Also, none of the studies on IDC currently contain whole-body data of both CT and MR: https://portal.imaging.datacommons.cancer.gov/explore/filters/?Modality_op=AND&Modality=PT&Modality=CT&tcia_tumorLocation=Whole%20Body.

fedorov: > and subsequently only use the resampling part of the TCIA codebase

I do not see how it is justified to adopt this random package just for the sake of resampling. Resampling is just a few lines of SimpleITK. Plastimatch has a resampling operator: https://plastimatch.org/plastimatch.html#plastimatch-resample.

fedorov: > I agree, but the model was trained and validated on the [(FDG-PET-CT-Lesions)](https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=93258287) dataset specifically, the model might not generalize that well to other datasets.

If the model does not generalize well to other datasets, and we know about it, then I do not see the point of adding this model. And I forgot that this collection you mention is private. Unless we have a publicly available sample in IDC for this model, I do not see how we can add it to MHub.

I understand that in the year that has already passed you had a contract that allowed data from sources other than IDC. In the current year, the requirement will be that all models must be accompanied by a sample from IDC that can be used to demonstrate that the model is functional.

silvandeleemput: > > and subsequently only use the resampling part of the TCIA codebase
> 
> I do not see how it is justified to adopt this random package just for the sake of resampling. Resampling is just a few lines of SimpleITK. Plastimatch has a resampling operator: https://plastimatch.org/plastimatch.html#plastimatch-resample.

If you read the lines below that line carefully, you'll see that I am actually open to removing the dependency on this TCIA library. Indeed resampling is available in other packages, I just want it to be similar to the one used for the model data in the GC challenge to allow more reproducible results. If these other packages resample in the exact the same way I am fine with that too.

silvandeleemput:
@LennyN95 I found two MHub-related issues trying to convert the DICOM files to Nifti using the following pipeline setup:

```
execute:
- FileStructureImporter
- NiftiConverter

modules:
  FileStructureImporter:
    input_dir: 'input_data'
    structures:
      - $sid@instance/re:.*::/re:.*-GK.*::@dicom:mod=ct
      - $sid/re:.*::/re:.*-PET.*::@dicom:mod=pt
    import_id: sid

  NiftiConverter:
    engine: dcm2niix
    allow_multi_input: true
    converted_file_name: '[basename].nii.gz'
```

The first structure is the CT DICOM image and the second structure is the PT DICOM image both belonging to the same instance `sid`. Both should be converted from Dicom to Nifti before subsequent processing.


1. First of all the `dcm2niix` engine fails to convert the DICOM input generating a lot of warnings and errors like the following and it won't generate a Nifti as output:
```
Warning: Interslice distance varies in this volume (incompatible with NIfTI format).

Warning: Missing images? Expected 3 images, but instance number (0020,0013) ranges from 42 to 221

Error: Too many NIFTI images with the name /app/data/imported_instances/PETCT_0011f3deaf/nifti/4.000000-GK p.v.3-7858
1_e1
```
I tested if this could be caused by having unsorted DICOM data using the DicomImporter with the sort option enabled instead of the FileStructureImporter but it gave me the same error. Could it be that the tool isn't correctly called for this use case? When changing it to `plastimatch` as a backend it seems to work.

2. When changing the engine of the NiftiConverter to the `plastimatch` engine the conversion seems to work, but then only one of the two images is converted (CT only) and the other isn't. Both are required for the model in subsequent steps to work. Why does this happen? Should I change my structure definitions? Using the debug log I get the following structure after the FileStructureImporter (which seems correct to me):
```

. Instance  [/app/data/_global]
├── id: global
├── sid: global
. Instance  [/app/data/imported_instances/PETCT_0011f3deaf]
├── id: bbe5fd9d-6486-4c1c-8b28-22fa10ab4519
├── _ref: PETCT_0011f3deaf
├── sid: PETCT_0011f3deaf
├── DICOM [/app/data/input_data/PETCT_0011f3deaf/03-23-2003-NA-PET-CT Ganzkoerper  primaer mit KM-10445/7.000000-PET corr.-19181] ✓
│   ├── mod: pt
│   └── sid: PETCT_0011f3deaf
├── DICOM [/app/data/input_data/PETCT_0011f3deaf/03-23-2003-NA-PET-CT Ganzkoerper  primaer mit KM-10445/4.000000-GK p.v.3-78581] ✓
│   ├── mod: ct
│   └── sid: PETCT_0011f3deaf
```

And this is the output after running the NiftiConverter with `plastimatch` as engine:
```
. Instance  [/app/data/_global]
├── id: global
├── sid: global
. Instance  [/app/data/imported_instances/PETCT_0011f3deaf]
├── id: bbe5fd9d-6486-4c1c-8b28-22fa10ab4519
├── _ref: PETCT_0011f3deaf
├── sid: PETCT_0011f3deaf
├── DICOM [/app/data/input_data/PETCT_0011f3deaf/03-23-2003-NA-PET-CT Ganzkoerper  primaer mit KM-10445/7.000000-PET corr.-19181] ✓
│   ├── mod: pt
│   └── sid: PETCT_0011f3deaf
├── DICOM [/app/data/input_data/PETCT_0011f3deaf/03-23-2003-NA-PET-CT Ganzkoerper  primaer mit KM-10445/4.000000-GK p.v.3-78581] ✓
│   ├── mod: ct
│   └── sid: PETCT_0011f3deaf
├── NIFTI [/app/data/imported_instances/PETCT_0011f3deaf/nifti/4.000000-GK p.v.3-78581.nii.gz] ✓
│   ├── mod: ct
│   └── sid: PETCT_0011f3deaf
├── LOG [/app/data/imported_instances/PETCT_0011f3deaf/nifti/4.000000-GK p.v.3-78581.pmconv.log] ✓
│   ├── mod: ct
│   ├── sid: PETCT_0011f3deaf
│   └── log-task: conversion

```

fedorov:
@silvandeleemput if the sample you are using is from IDC, please provide the details. If it is not from IDC, it is not suitable for debugging the issues you are describing. It is indeed important to debug on data outside of IDC, but we first need to get it to work on a sample from IDC.

silvandeleemput:
Fair enough, I wasn't using IDC data but the TCIA requiring a license. However, using the following test IDC DICOM set: [Anti-PD-1_Lung](https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=41517500) with Patient ID: PD-1-Lung-00001, the same issues occur. The only difference is that the `dcm2niix` tool generates less errors and warnings, but still doesn't produce a Nifti. New pipeline config file:

```
execute:
- FileStructureImporter
- NiftiConverter

modules:
  FileStructureImporter:
    input_dir: 'input_data'
    structures:
      - $sid@instance/re:.*::/re:.*-CT Lung.*::@dicom:mod=ct
      - $sid/re:.*::/re:.*-PET NAC.*::@dicom:mod=pt
    import_id: sid

  NiftiConverter:
    engine: dcm2niix
    allow_multi_input: true
    converted_file_name: '[basename].nii.gz'
```

LennyN95:
You need to adjust the `in_datas` argument of the NiftiConverter. By default, we only look for `CT` data ([as can be seen in our documentation here](https://github.com/MHubAI/documentation/blob/main/documentation/mhubio/mhubio_modules.md#nifticonverter)).
This is questionable and will be changed in the near future (the discussion is whether a default value makes sense given the variety of possible use cases. But whether or not, CT shouldn't be preferred over e.g. MR. But allowing the converter to convert everything by default seems just as wrong).

Please change the current default from
` in_datas: dicom|nrrd|mha:mod=ct`
to
`in_datas: dicom|nrrd|mha:mod=ct|pt`.

silvandeleemput:
> You need to adjust the `in_datas` argument of the NiftiConverter. By default, we only look for `CT` data ([as can be seen in our documentation here](https://github.com/MHubAI/documentation/blob/main/documentation/mhubio/mhubio_modules.md#nifticonverter)).
> This is questionable and will be changed in the near future (the discussion is whether a default value makes sense given the variety of possible use cases. But whether or not, CT shouldn't be preferred over e.g. MR. But allowing the converter to convert everything by default seems just as wrong).

That was indeed the issue here. I agree that the current setting for the CT modality is very counterintuitive. I kind of expected it to convert all images (or dicom at least) instead like the MhaConverter does. I think it is at least desirable to drop to modality check in the filter. Whatever you decide here, it might be nice to also display the value of the `in_datas` filter for the converter modules when running in debug mode to notify the developer that such a filter is being applied.

LennyN95:
Great suggestion @silvandeleemput!

fedorov:
> But allowing the converter to convert everything by default seems just as wrong

I am not sure this is the case. I would want to understand better how this works right now. Maybe you can step me through the process @LennyN95 at one of our regular meetings.

I do not think it is the job of MHub wrapper to enforce the expectations of the model. Or at least there should be an easy way to disable those. If the model expects CT, but I want to try it out on MR, I do not see why wrapper should forbid me from doing that.

LennyN95:
While I agree, both is actually possible and depends on how the ModelRunner is implemented. You can explicitly filter only for specific data or you can make it configurable. 

This makes it transparent from the code header (just look at the python input decorators) what a Module is able to take. 

Happy to demonstrate in detail next week ;)

models/gc_autopet_fpr/config/default.yml
fedorov: What is `dicom:mod=ct:type=GK and dicom:mod=ct:type=PET`? There is not PET type of DICOM CT modality, and I have no idea what GK is?!

fedorov: Per model metadata, the inputs are DICOM CT and DICOM PET, see https://github.com/MHubAI/models/pull/52/files#diff-dba8e5743cedb6dd6171269267ec35e5ae16927b3e8232afea60fc4472b40743R7.

LennyN95: @fedorov is right here, please refer to my first example above where I proposed a solution for this specific case. The second example is more of a general technical possibilit, hence the explicit notion of not beeing suitable for this model (_This won't fit the described use case as we've two different modalities [...]_).

silvandeleemput: The configuration for the FileStructureImporter was configured to work out of the box for the data structure from the [FDG-PET-CT-Lesions](https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=93258287) dataset. 

This of course won't work in general for other datasets, so a general structured interface like you propose is a better alternative. I propose to change the GK into CT though, because GK was specific to the FDG-PET-CT-Lesions dataset and does not make any sense out of context.

I think indeed that the first example is the clearer solution in this specific case. The structure filter will become:

```
structures:
- $sid@instance/CT@dicom:mod=ct
- $sid/PET@dicom:mod=pet
```

models/gc_autopet_fpr/utils/TciaDicomProcessor.py
silvandeleemput: Yes, we can, we can also use other alternatives like nilearn or SimpleITK, the actual resampling code used in the Tcia preprocessing code is:
```
nilearn.image.resample_to_img(ct, pet, fill_value=-1024)
```

So I am for dropping the Tcia-specific preprocessing package and running either the plastimatch resampling or the nilearn resampling. I should be able to mimic the nilearn instruction in plastimatch resample using `--default-value -1024`. Do you have a preference?

LennyN95: Plastimatch is preferred whenever applicable.

silvandeleemput: In that case, what should I pass on for the log file `path_to_log_file=log_data.abspath,`?

fedorov:
As discussed today, the model being added here is this one: https://grand-challenge.org/algorithms/multi_stage_hybrid_cnn_13/

models/gc_autopet_fpr/utils/TciaDicomProcessor.py
LennyN95: Just use an `@IO.Output` decorator for the log file. One example (which is multi-out though, but still comparable) can be found in our NiftiConverter: 

https://github.com/MHubAI/mhubio/blob/4c31c53a843f35f317a67254dffe8d4eae7e9881/mhubio/modules/convert/NiftiConverter.py#L110

silvandeleemput: The resampling has been implemented using plastimatch.

silvandeleemput:
[output_gc_autopet_fpr.zip](https://github.com/MHubAI/models/files/15109310/output_gc_autopet_fpr.zip)

silvandeleemput:
/test

```yaml
sample:
  idc_version: 17.0
  data:
    - SeriesInstanceUID: 1.3.6.1.4.1.14519.5.2.1.7009.2403.798861112144623421423086363370
      aws_url: s3://idc-open-data-two/661c29d5-0b8e-4016-a119-58d461d86810/*
      path: case1/ct
    - SeriesInstanceUID: 1.3.6.1.4.1.14519.5.2.1.7009.2403.291916523874874486349020167447
      aws_url: s3://idc-open-data-two/3995efef-beef-4f2e-9782-f44700a1064c/*
      path: case1/pet

reference:
  url: https://github.com/MHubAI/models/files/15109310/output_gc_autopet_fpr.zip
```

---

<details><summary>Test Results (24.04.25_14.08.51_xHF1bsAg8N)</summary>

```yaml
id: 7c2fffcc-3f48-4f7e-bdb8-ff546f3b9022
date: '2024-04-25 14:16:18'
checked_files:
- file: autopet_fpr_tumor_segmentation.seg.dcm
  path: /app/test/src/case1/autopet_fpr_tumor_segmentation.seg.dcm
  checks:
  - checker: DicomsegContentCheck
    notes:
    - label: Segment Count
      description: The number of segments identified in the inspected dicomseg file.
      info: 1
- file: autopet_fpr_tumor_segmentation.mha
  path: /app/test/src/case1/autopet_fpr_tumor_segmentation.mha
  checks:
  - checker: ImageFileCheck
    notes:
    - label: Data Type
      info: uint8
    - label: Dice Score
      info: 0.999779
    findings:
    - label: Dice Score Difference
      description: Dice score between reference and test image exceeds tolerance.
      info:
        dice_score: 0.999779
        tolerance: 0.0001
summary:
  files_missing: 0
  files_extra: 0
  checks:
    DicomsegContentCheck:
      files: 1
    ImageFileCheck:
      files: 1
      findings:
        Dice Score Difference: 1
conclusion: false
```

</details>

---

<details><summary>Test Results (24.06.03_10.26.45_9t3n5BDtdf)</summary>

```yaml
id: 6daad041-74d3-4514-9bd0-ce662026a70a
date: '2024-06-03 10:36:14'
checked_files:
- file: autopet_fpr_tumor_segmentation.seg.dcm
  path: /app/test/src/case1/autopet_fpr_tumor_segmentation.seg.dcm
  checks:
  - checker: DicomsegContentCheck
    notes:
    - label: Segment Count
      description: The number of segments identified in the inspected dicomseg file.
      info: 1
- file: autopet_fpr_tumor_segmentation.mha
  path: /app/test/src/case1/autopet_fpr_tumor_segmentation.mha
  checks:
  - checker: ImageFileCheck
    notes:
    - label: Data Type
      info: uint8
    - label: Dice Score
      info: 0.999779
    findings:
    - label: Dice Score Difference
      description: Dice score between reference and test image exceeds tolerance.
      info:
        dice_score: 0.999779
        tolerance: 0.0001
summary:
  files_missing: 0
  files_extra: 0
  checks:
    DicomsegContentCheck:
      files: 1
    ImageFileCheck:
      files: 1
      findings:
        Dice Score Difference: 1
conclusion: false
```

</details>

silvandeleemput:
/review

This algorithm/model is now ready for review. 

Notable updates:
* We found a suitable dataset on IDC (ACRIN-NSCLC-FDG-PET) with plenty of PET/CT pairs, our test example is drawn from this dataset
* The input structure filter has been changed to:
```
structures:
- $sid@instance/ct@dicom:mod=ct
- $sid/pet@dicom:mod=pet
```
* The resampling of the CT to the PET is now performed using plastimatch
* Additional resampling is performed to ensure the model can run on any PET/CT inputs, while keeping the output segmentations in the same physical space as the input images.
* We added a CLI to prevent nnUNet output clutter
* The meta.json has been updated and actualized

fedorov:
> Corrected winner to runner up after contact with the challenge organizers

@miriam-groeneveld what happened with the winner, why are we including the runner-up instead of the winner?

> https://grand-challenge.org/algorithms/multi_stage_hybrid_cnn_13

This page is not publicly available. Is the URL correct?

miriam-groeneveld:
> > Corrected winner to runner up after contact with the challenge organizers
> 
> @miriam-groeneveld what happened with the winner, why are we including the runner-up instead of the winner?

The winner's algorithm does not have the weights publicly available, so wasn't suited.

> > https://grand-challenge.org/algorithms/multi_stage_hybrid_cnn_13
> 
> This page is not publicly available. Is the URL correct?

Yes, the algorithm is private. We have contacted the owners of the algorithm with the request to make it public, but haven't received a response. We can remove the link to the GC algorithm and only link to the challenge.

models/gc_autopet_fpr/config/default.yml
```diff
@@ -0,0 +1,39 @@
...
+  NiftiConverter:
+    engine: plastimatch
+    allow_multi_input: true
+    converted_file_name: '[basename].nii.gz'
+    in_datas: dicom:mod=ct|pt
+
+  DsegConverter:
+    model_name: GCAutoPETfpr
+    target_dicom: dicom:mod=pt
+    source_segs: [mha:mod=seg:roi=*]
```
LennyN95: Needs to be a string.

LennyN95 (changes requested):

models/gc_autopet_fpr/config/default.yml
silvandeleemput: Requested change has been implemented.

silvandeleemput:
@LennyN95 Regarding the results, some differences between the reference and test image segmentation are expected due to hardware/software GPU differences, the found overlap Dice score (0.999779) seems acceptable.

LennyN95 (approved):

PR #53, opened 2023-09-05
MHub / GC - Add SPIDER baseline model/algorithm

silvandeleemput:
This PR adds the SPIDER ISS baseline model (from the [SPIDER challenge](https://spider.grand-challenge.org/)) to MHub.
GitHub Repo: https://github.com/DIAGNijmegen/SPIDER-Baseline-IIS
GC page: https://grand-challenge.org/algorithms/spider-baseline-iis/

### Algorithm I/O
  * Input:
    * Sagittal spine MR image (DICOM)
  * Output:
    * Sagittal spine MR segmentation  (0: background, 1-25: different vertebrae numbered from the bottom (i.e. L5 = 1) 100: spinal canal 101-125 for partially visible vertebrae 201-225: different intervertebral discs (also numbered from the bottom, i.e. L5/S1 = 201))
    * training data set: https://zenodo.org/record/8009680 (MHA)
    * open-source DICOM data set: https://www.cg.informatik.uni-siegen.de/en/spine-segmentation-and-analysis

### Caveats
* The PR target is `main`, but should be something like `m-gc-spider-baseline`
* ~Support for Dicom input has not been added for this algorithm since we could not find any specific Spine MR data on IDC, if it exists let us know then we can add it.~ Support has been added.
* ~Support for DicomSeg output has not been added for this algorithm because of the complexity and sheer number of output segmentation labels.~ Basic support for DicomSeg has been added with support for 1-25: different vertebrae numbered from the bottom (i.e. L5 = 1) through the dseg rois.


models/gc_spider_baseline/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# TODO Clone the main branch of MHubAI/models
+#RUN git stash \
+#    && git fetch https://github.com/MHubAI/models.git main \
+#    && git merge FETCH_HEAD \
+#    && git sparse-checkout set "models/gc_spider_baseline" \
+#    && git fetch https://github.com/MHubAI/models.git main
+
+# Add two static configuration json files required by the SPIDER baseline algorithm
+RUN echo '{"filters": 50, "traversal_direction": "up"}' > /app/src/experiments/SPIDER-Baseline/arguments.json
```
LennyN95: Hey Sil, would there be any use case where one want to derive from these static parameters? 

If yes, I'd suggest we expose (some of) them via @IO.Config and construct the json at Module initialization.

In general, could we move these as files into `/app/models/$model_name/src`? The current implementation is totally valid; however it won't scale well for large configs.

models/gc_spider_baseline/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN echo '[{\
+    "identifier": "input_img",\
+    "subset": "testing",\
+    "filetype": "mha",\
+    "slice-order-reversed": false,\
+    "rescale-intercept": 0,\
+    "modality": "MRI",\
```
LennyN95: E.g., from the paper it reads like the algorithm supports multiple modalities; 

Out DicomImporter automatically reads the modality from the Dicom headers when run with default configuration; so instead I'd use a variable input (e.g., `mha:mod=ct|mr`) and then check the actual modality in the model runner's `task()` method and set the config from there.

LennyN95:
Hi Sil, 

> Support for DicomSeg output has not been added for this algorithm because of the complexity and sheer number of output segmentation labels.

We need to discuss this, ideally with @fedorov. 

The vertebrae (1-25) should all be found in SegDB. For the partially vertebrae, maybe there is some kind of modifier to indicate incompleteness, Andrey will know this best. 

SCT has codes for the intervertebral discs, so this shouldn't be a problem. 

So i think this is solvable; if some variable delineations cannot be covered by Dicom for any reason, we might exclude them and only predict those that are (we can always export additional data). But for those we can standardize, we need to. After all,a Dicom-Dicom default pipeline is a hard requirement.

Best,
Leo.

fedorov:
> Sagittal spine MR image (MHA) [...] Support for Dicom input has not been added for this algorithm since we could not find any specific Spine MR data on IDC, if it exists let us know then we can add it.

My understanding is that availability of the applicable data in IDC was precisely one of the criteria used to select the models for the contract. So I assume you did find it at the time the selection was made, right?

> Sagittal spine MR segmentation (0: background, 1-25: different vertebrae numbered from the bottom (i.e. L5 = 1) 100: spinal canal 101-125 for partially visible vertebrae 201-225: different intervertebral discs (also numbered from the bottom, i.e. L5/S1 = 201))

Individual vertebrae are already mapped into SNOMED in https://github.com/MHubAI/models/blob/main/models/totalsegmentator/config/dicomseg_metadata_whole.json, so we can just reuse those.

For the spinal canal, is there an anatomic designation of spinal canal segments corresponding to the individual vertebrae? If not, I am not sure if it makes sense to have a vertebrae-labeled canal segmentation. Also, I am confused about "partially visible vertebrae". What is the reasoning for the logic: "if vertebra is partial, we segment spinal canal"?

image

For the disks, there are codes to leverage from SNOMED (https://browser.ihtsdotools.org/?perspective=full&conceptId1=404684003&edition=MAIN/2023-09-01&release=&languages=en):

image

silvandeleemput:
> My understanding is that availability of the applicable data in IDC was precisely one of the criteria used to select the models for the contract. So I assume you did find it at the time the selection was made, right?

Yes, there is some spine MR available on IDC: 
* [CMB-CRC](https://portal.imaging.datacommons.cancer.gov/explore/filters/?collection_id=cmb_crc)
* [CMB-PCA](https://portal.imaging.datacommons.cancer.gov/explore/filters/?collection_id=cmb-pca)

However, we found out when adding the model that these are not very suited for running with the model. There is however a very suitable open-source DICOM dataset available outside of IDC here: https://www.cg.informatik.uni-siegen.de/en/spine-segmentation-and-analysis. Hence, I have re-enabled DICOM support for the inputs for this model.

Thanks for the detailed instructions regarding the DicomSeg labels. I think it should be possible to add these, but I understood that it is required for the segmentation labels to be presented in sequence without any gaps in order for the DicomSegConverter to properly work. Wouldn't this require us to remap the internal labels to be without any gaps? 

> Also, I am confused about "partially visible vertebrae". What is the reasoning for the logic: "if vertebra is partial, we segment spinal canal"?

I am not entirely sure I fully understand the idea for this from the algorithm authors, but it seems to be something like you mentioned; if the vertebrae is on the edge of the image and not fully visible within the scan (so partially visible) label it and start searching for the next segment label it, etc...

fedorov:
> However, we found out when adding the model that these are not very suited for running with the model. 

Can you please elaborate on this?

silvandeleemput:
> Can you please elaborate on this?

Yes, we tried running the algorithm on some of the MR spine IDC data, but the algorithm failed to segment any spine segments, resulting in failing and empty output segmentations. Upon inspection, the MR images found in the IDC data have fewer slices and a much higher slice thickness and also have more surrounding organ tissue than the scans the model was trained on which seems to be causing the issue. The other DICOM dataset that I mentioned contains MR Spine images with properties closer to the original training data set and returns good output segmentations.

silvandeleemput:
Basic support for DicomSeg has been added with support for 1-25: different vertebrae numbered from the bottom (i.e. L5 = 1) through the dseg rois. Currently, the partially visible vertebrae labels have just been remapped to the normal vertebrae labels, and the spinal canal and intervertebral discs are ignored. 

The implementation was achieved through using the DsegConverter using the already available dseg rois. If the other segmentation rois get added it should be very easy to add the other segmentation outputs as well (remapping code is already available). If you guys prefer a dseg.json setup instead let me know then I can generate that as well.

fedorov:
Here's a better link that has discs codes: https://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_7604.html

silvandeleemput:
Could you please confirm that you guys prefer a dseg.json including the disks and spinal canal over the current implementation using the available values in https://github.com/MHubAI/SegDB (only vertebrae)?

If so could someone help with the DicomSeg codes for the spinal canal as well?

LennyN95:
> Could you please confirm that you guys prefer a dseg.json including the disks and spinal canal over the current implementation using the available values in https://github.com/MHubAI/SegDB (only vertebrae)?

No, we're always preferring the usage of SegDB codes ;) Sorry if that wasn't clear; The `dseg.json` is only required if SegDB entries are missing and will ultimately be integrated into SegDB. In that case, we'll update the implementation to use SegDB instead. 

@fedorov commented earlier on the spinal canal codes. @Sil, the problem is that you have multiple segments of the spinal canal, correct? Couldn't we assign the spinal canal SCT code to all these segments for now @fedorov? We can always update with more specific codes in the future. 

A nice feature of MHub is, that besides the standardized output, all original outputs of the pipeline are still available. So we can always have a workflow that returns those output files.

fedorov:
> the problem is that you have multiple segments of the spinal canal, correct? Couldn't we assign the spinal canal SCT code to all these segments for now @fedorov? We can always update with more specific codes in the future.

You mean assign generic "Spinal canal" code instead of more specific ones?

silvandeleemput:
> > Could you please confirm that you guys prefer a dseg.json including the disks and spinal canal over the current implementation using the available values in https://github.com/MHubAI/SegDB (only vertebrae)?
> 
> No, we're always preferring the usage of SegDB codes ;) Sorry if that wasn't clear; The `dseg.json` is only required if SegDB entries are missing and will ultimately be integrated into SegDB. In that case, we'll update the implementation to use SegDB instead.
> 
Thanks!

> @fedorov commented earlier on the spinal canal codes. @sil, the problem is that you have multiple segments of the spinal canal, correct? Couldn't we assign the spinal canal SCT code to all these segments for now @fedorov? We can always update with more specific codes in the future.
> 

That's exactly the implementation in place right now, and I am using the current spinal canal SCT codes available in the SegDB.
However, the algorithm provides more segmentation labels, so the codes that are missing in SegDB are:
* Vertebrae disc codes (see link of Fedorov: https://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_7604.html)
* A generic "Spinal canal" code

I think it would be great if you could add these to the SegDB.

> A nice feature of MHub is, that besides the standardized output, all original outputs of the pipeline are still available. So we can always have a workflow that returns those output files.

Yes, this is currently the case, I also output the original MHA segmentation alongside the DicomSeg output.

silvandeleemput:
> You mean assign generic "Spinal canal" code instead of more specific ones?

Yes, I don't think this algorithm/model is very specific and just segments it as a whole.

fedorov:
> Vertebrae disc codes (see link of Federov:

Fedorov

> I don't think this algorithm/model is very specific and just segments it as a whole.

You can use "Spinal canal" SCT=61853006 https://browser.ihtsdotools.org/?perspective=full&conceptId1=61853006&edition=MAIN/2023-11-01&release=&languages=en

models/gc_spider_baseline/dockerfiles/Dockerfile
silvandeleemput: > Hey Sil, would there be any use case where one want to derive from these static parameters?
> 
maybe the `traversal_direction`, filters should remain 50 to match the model

> If yes, I'd suggest we expose (some of) them via @IO.Config and construct the json at Module initialization.
> 
fine, the `traversal_direction` can be exposed.

> In general, could we move these as files into `/app/models/$model_name/src`? The current implementation is totally valid; however it won't scale well for large configs.

It should be possible to do so. I will not do this for the current algorithms, because it is quite some work and falls beyond the scope of the two months earlier established quality checklist. If you want to do so feel free to do it.

models/gc_spider_baseline/dockerfiles/Dockerfile
silvandeleemput: Ok, support for CT has been added

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
```diff
@@ -0,0 +1,155 @@
...
+        'spider_baseline_vertebrae_segmentation.mha',
+        'mha:mod=seg:model=SpiderBaselineSegmentation:seg=remapped:roi=VERTEBRAE_L5,VERTEBRAE_L4,VERTEBRAE_L3,VERTEBRAE_L2,VERTEBRAE_L1,VERTEBRAE_T12,VERTEBRAE_T11,VERTEBRAE_T10,VERTEBRAE_T9,VERTEBRAE_T8,VERTEBRAE_T7,VERTEBRAE_T6,VERTEBRAE_T5,VERTEBRAE_T4,VERTEBRAE_T3,VERTEBRAE_T2,VERTEBRAE_T1,VERTEBRAE_C7,VERTEBRAE_C6,VERTEBRAE_C5,VERTEBRAE_C4,VERTEBRAE_C3,VERTEBRAE_C2,VERTEBRAE_C1',
+        data='in_data',
+        bundle='model',
+        the='Spider baseline vertebrae segmentation for the input sagittal spine MRI. Remapped segmentation output: '
+            '0: background '
+            '1-25: different vertebrae numbered from the bottom (i.e. L5 = 1) (includes partially visible vertebrae from raw segmentation)'
+    )
+    def task(self, instance: Instance, in_data: InstanceData, out_data_raw: InstanceData, out_data: InstanceData) -> None:
+        in_data_path = Path(in_data.abspath)
```
LennyN95: Does the `Path` wrapping happen for any reason? If no, please remove as it decreases the readablity.

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
```diff
@@ -0,0 +1,155 @@
...
+        internal_img_path = self.SPIDER_INTERNAL_DATASET_PATH / "images" / "input_img.mha"
+        self.log(f"Copying data files to internal SPIDER data structure: {in_data_path} -> {internal_img_path}", level="NOTICE")
+        shutil.copy(str(in_data_path), str(internal_img_path))
+
+        filetype = str(in_data.type.ftype).lower()
+        modality = in_data.type.meta["mod"].upper()
```
LennyN95: To compare metadata please use the following syntax:

```Python
in_data.type.meta <= Meta(mod='mr')
in_data.type.meta <= MR # ... as we have a shortcut for modalities
```

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
```diff
@@ -0,0 +1,155 @@
...
+        is_modality_mr = modality == "MR"
+
+        if modality not in ["CT", "MR"]:
```
LennyN95: This makes the code hard to follow.
I suggest two methods, `process_mr` and `process_ct`, that handle the two cases or two set-up methods, e.g., `setup_mr` and `setup_ct`, that set up the algorithm accordingly. 

Then, the task method can be cleaned up and organized into three sections:
- check modality
- setup algorithm
- run algorithm

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
```diff
@@ -0,0 +1,155 @@
...
+            '0: background '
+            '1-25: different vertebrae numbered from the bottom (i.e. L5 = 1) (includes partially visible vertebrae from raw segmentation)'
+    )
+    def task(self, instance: Instance, in_data: InstanceData, out_data_raw: InstanceData, out_data: InstanceData) -> None:
+        in_data_path = Path(in_data.abspath)
+        internal_img_path = self.SPIDER_INTERNAL_DATASET_PATH / "images" / "input_img.mha"
+        self.log(f"Copying data files to internal SPIDER data structure: {in_data_path} -> {internal_img_path}", level="NOTICE")
+        shutil.copy(str(in_data_path), str(internal_img_path))
+
+        filetype = str(in_data.type.ftype).lower()
```
LennyN95: This might fail if we ever change the string representation. Also, as this is passed directly into the generated config file, does the algorithm really support every possible file type?

Instead, map MHub-IO file types towards what the algorithm expects:
```Python
if in_data.type.ftype == FileType.MHA: 
   filetype = "mha"
elif [..]:
else: 
   raise ValueError(f"Unsupported file type {in_data.type.ftype}.") 
   
```

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
```diff
@@ -0,0 +1,155 @@
...
+from TestSpine import Module as SpiderAlgorithm
+
+
+@IO.Config('traversal_direction_up', bool, True, the="Direction to traverse the image")
+@IO.ConfigInput('in_data', 'nifti|nrrd|mha:mod=ct|mr', the='supported datatypes for the spider baseline model')
+class SpiderBaselineRunner(Module):
+
+    traversal_direction_up: bool
+
+    SPIDER_DATA_DIR = Path(os.environ["VERSEG_BASEDIR"])
```
LennyN95: Why don't we set these as file globals so that we can omit the `self.` ?
They're not MHub-specific, and moving them outside of the Module class can improve readability.

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def create_arguments_config_file(self, traversal_direction_up: bool = True):
+        with open(self.SPIDER_EXPERIMENT_ARGUMENTS_FILE, "w") as f:
+            json.dump({
+                "filters": 50,  # Fixed for this model
+                "traversal_direction": "up" if traversal_direction_up else "down"
+                }, f
+            )
+
+    def create_input_metadata_file(self, modality: str, filetype: str = "mha"):
```
LennyN95: Please explain what this is doing. References to the original algorithm's documentation can be helpful too.

models/gc_spider_baseline/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+#    && git merge FETCH_HEAD \
+#    && git sparse-checkout set "models/gc_spider_baseline" \
+#    && git fetch https://github.com/MHubAI/models.git main
+
+# Setup required environment variables for SPIDER algorithm
+ENV PYTHONPATH="/app:/app/src/devel"
+ENV VERSEG_BASEDIR="/app/src/"
+
+# Default run script
+WORKDIR "/app"
```
LennyN95: WD has been set to `/app` in the base image already.

models/gc_spider_baseline/dockerfiles/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM mhubai/base:latest
+
+# Update authors label
+LABEL authors="sil.vandeleemput@radboudumc.nl"
+
+# Install PyTorch 2.0.1 (CUDA enabled)
+RUN pip3 install --no-cache-dir torch==2.0.1+cu118 -f https://download.pytorch.org/whl/torch_stable.html
+
+# Install git-lfs (required for unpacking model weights)
+RUN apt update && apt install -y --no-install-recommends git-lfs && rm -rf /var/lib/apt/lists/*
```
LennyN95: For consistency, let's have one line per command.

models/gc_spider_baseline/config/default.yml
```diff
@@ -0,0 +1,36 @@
...
+- DsegConverter
+- DataOrganizer
+
+modules:
+  DicomImporter:
+    source_dir: input_data
+    import_dir: sorted_data
+    sort_data: true
+    meta:
+      mod: mr
```
LennyN95: Doesn't the algorithm support CT and MR?
You can set `mod: '%Modality'` to extract the modality from the dicom header.

models/gc_spider_baseline/config/default.yml
```diff
@@ -0,0 +1,36 @@
...
+
+  MhaConverter:
+    engine: panimg
+
+  DsegConverter:
+    model_name: GCSpiderBaseline
+    target_dicom: dicom:mod=mr
```
LennyN95: Also, limiting to MR only.

models/gc_spider_baseline/dockerfiles/Dockerfile
LennyN95: >  If you want to do so feel free to do it.

??

models/gc_spider_baseline/dockerfiles/Dockerfile
LennyN95: Great, the default config is fixed on MR, though (I added some comments and suggestions there).

LennyN95:
> You can use "Spinal canal" SCT=61853006 https://browser.ihtsdotools.org/?perspective=full&conceptId1=61853006&edition=MAIN/2023-11-01&release=&languages=en

I updated SegDB. 
The generic id for the spinal canal is `SPINAL_CANAL`.

silvandeleemput:
@LennyN95 Did you also add the Vertebrae discs to SegDB yet?
Vertebrae disc codes (see link of Fedorov: https://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_7604.html)

LennyN95:
Yup, I pushed the changes a second ago.
You can search for "disk" here: https://github.com/MHubAI/SegDB/blob/main/segdb/data/segmentations.csv

The SegDB IDs follow the schema `VERTEBRAE_DISK_C2C3`, `VERTEBRAE_DISK_C3C4`, ...

models/gc_spider_baseline/dockerfiles/Dockerfile
silvandeleemput: Removed

models/gc_spider_baseline/config/default.yml
silvandeleemput: Added

models/gc_spider_baseline/config/default.yml
silvandeleemput: Added CT

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
silvandeleemput: That's fine, I'll do that.

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
silvandeleemput: I will even simplify it further since MHA is the only accepted file type so the file type extraction can be removed altogether.

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
silvandeleemput: Ok

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
silvandeleemput: I'll remove it

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
silvandeleemput: Ok, I have significantly refactored the code based on your suggestions, please have a look.

models/gc_spider_baseline/dockerfiles/Dockerfile
silvandeleemput: These options have been added now.

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
silvandeleemput: I added some comments, but I haven't been able to find any documentation on this.

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
```diff
@@ -0,0 +1,195 @@
...
+        'VERTEBRAE_L1,VERTEBRAE_T12,VERTEBRAE_T11,VERTEBRAE_T10,VERTEBRAE_T9,VERTEBRAE_T8,VERTEBRAE_T7,VERTEBRAE_T6,VERTEBRAE_T5,'
+        'VERTEBRAE_T4,VERTEBRAE_T3,VERTEBRAE_T2,VERTEBRAE_T1,VERTEBRAE_C7,VERTEBRAE_C6,VERTEBRAE_C5,VERTEBRAE_C4,VERTEBRAE_C3,VERTEBRAE_C2,VERTEBRAE_C1,'
+        'VERTEBRAE_DISK_L5S1,VERTEBRAE_DISK_L4L5,VERTEBRAE_DISK_L3L4,VERTEBRAE_DISK_L2L3,VERTEBRAE_DISK_L1L2,VERTEBRAE_DISK_T12L1,VERTEBRAE_DISK_T11T12,'
+        'VERTEBRAE_DISK_T10T11,VERTEBRAE_DISK_T9T10,VERTEBRAE_DISK_T8T9,VERTEBRAE_DISK_T7T8,VERTEBRAE_DISK_T6T7,VERTEBRAE_DISK_T5T6,VERTEBRAE_DISK_T4T5,'
+        'VERTEBRAE_DISK_T3T4,VERTEBRAE_DISK_T2T3,VERTEBRAE_DISK_T1T2,VERTEBRAE_DISK_C7T1,VERTEBRAE_DISK_C6C7,VERTEBRAE_DISK_C5C6,VERTEBRAE_DISK_C4C5,'
+        'VERTEBRAE_DISK_C3C4,VERTEBRAE_DISK_C2C3,'
+        'SPINAL_CANAL',
+        data='in_data',
+        bundle='model',
+        the='Spider baseline vertebrae segmentation for the input sagittal spine MRI or CT. Remapped segmentation output: '
```
LennyN95, line 74: For better readability, this can go into the model card description (it's redundant here as we now have all SegDB IDs in the roi metadata).

silvandeleemput: I think you are referring to the index descriptions under `the`? I kind of understand your argument and I agree that it should be updated in the model card description, but it seems still nice to have a more readable summary of the rois section next to the roi list since large lists of roi strings are hard to read for users/developers.

LennyN95: I'm fine with both. To give you some context: I introduced the decorators not only to provide a standardized framework and increase readability but also because their easily parsable and will allow automatic generation of documentation for all (custom) MHub-IO modules in the future.

models/gc_spider_baseline/meta.json
```diff
@@ -0,0 +1,135 @@
...
+    },
+    "analyses": {
+      "title": "Evaluation",
+      "text": "The segmentation performance was evaluated using two metrics: (1) The Dice coefficient to measure the volume overlap, and (2) the average absolute surface distance (ASD) as an indication of the segmentation accuracy along the sur- face of all structures.",
+      "references": [],
+      "tables": []
+    },
+    "evaluation": {
+      "title": "Evaluation data",
+      "text": "The test set consists of 39 lumbar MRI studies of unique patients, which includes 15 out of the 20 fully manually annotated studies.",
```
LennyN95: Not clear to me, is studies = patients = n?
What are "the 20 fully manually annotated studies."?

models/gc_spider_baseline/meta.json
```diff
@@ -0,0 +1,135 @@
+{
+  "id": "aed5a82e-2188-4d50-83f9-45bce09f9daa",
+  "name": "spider_baseline",
+  "title": "SPIDER Baseline algorithm: Iterative Instance Segmentation of the spine",
+  "summary": {
+    "description": "The SPIDER Challenge focuses on the segmentation of three anatomical structures in lumbar spine MRI: vertebrae, intervertebral discs (IVDs), and spinal canal. The segmentation task requires similar algorithms to produce separate masks for each vertebra, IVD, and the spinal canal in the lumbar spine MRI volume. The vertebrae are subsequently labeled between 1 and 24, with 24 being the lowest lumbar vertebra (L5). Vertebra which are not completely visible within the field of view of the MR image are labeled between 101 and 124. The intervertebral discs are labeled between 201 and 224 and are decided by the vertebra above it. The spinal canal is labeled as 100.",
```
LennyN95: Also, here, this reads more like an introduction to the challenge, whereas we want to introduce a specific model (you can mention that it originates from the challenge).

models/gc_spider_baseline/meta.json
```diff
@@ -0,0 +1,135 @@
...
+    ],
+    "github": "https://github.com/DIAGNijmegen/SPIDER-Baseline-IIS",
+    "zenodo": "",
+    "colab": "",
+    "slicer": false
+  },
+  "info": {
+    "use": {
+      "title": "Intended use",
+      "text": "Segmentation of three anatomical structures in lumbar spine MRI: vertebrae, intervertebral discs (IVDs), and spinal canal.",
```
LennyN95: This describes what the algorithm is doing (redundand to description) not how it is intended to be used.

models/gc_spider_baseline/meta.json
```diff
@@ -0,0 +1,135 @@
...
+      "references": [],
+      "tables": []
+    },
+    "analyses": {
+      "title": "Evaluation",
+      "text": "The segmentation performance was evaluated using two metrics: (1) The Dice coefficient to measure the volume overlap, and (2) the average absolute surface distance (ASD) as an indication of the segmentation accuracy along the sur- face of all structures.",
```
LennyN95: Where are these evaluation results reported? Why don't we add them to the table below?

models/gc_spider_baseline/meta.json
```diff
@@ -0,0 +1,135 @@
...
+    },
+    "evaluation": {
+      "title": "Evaluation data",
+      "text": "The test set consists of 39 lumbar MRI studies of unique patients, which includes 15 out of the 20 fully manually annotated studies.",
+      "references": [],
+      "tables": []
+    },
+    "training": {
+      "title": "Training data",
+      "text": "This is a large publicly available multi-center lumbar spine magnetic resonance imaging (MRI) dataset with reference segmentations of vertebrae, intervertebral discs (IVDs), and spinal canal. The dataset includes 447 sagittal T1 and T2 MRI series from 218 studies of 218 patients with a history of low back pain. ",
```
LennyN95: How were the 179 training samples selected?

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
```diff
@@ -0,0 +1,194 @@
...
+
+SPIDER_DATA_DIR = Path(os.environ["VERSEG_BASEDIR"])
+SPIDER_INTERNAL_DATASET_PATH = SPIDER_DATA_DIR / "datasets" / "spider_input"
+SPIDER_EXPERIMENT_DIR = SPIDER_DATA_DIR / "experiments" / "SPIDER-Baseline"
+SPIDER_INTERNAL_OUTPUT_DIR = SPIDER_EXPERIMENT_DIR / "results" / "spider_input"
+SPIDER_EXPERIMENT_ARGUMENTS_FILE = SPIDER_EXPERIMENT_DIR / "arguments.json"
+SPIDER_INTERNAL_DATASET_META_FILE = SPIDER_INTERNAL_DATASET_PATH / "metadata.json"
+
+
+@IO.Config('traversal_direction_up', bool, True, the="Direction to traverse the image")
```
LennyN95: Some questions:

- Why does this parameter matter? 
- What happens if I set it to False? 
- Will the algorithm produce a different output?

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
```diff
@@ -0,0 +1,194 @@
...
+            raise ValueError(f"SpiderBaselineRunner does not support other modalities than CT and MR")
+        self.process(
+            in_data=in_data,
+            out_data_raw=out_data_raw,
+            **config
+        )
+        self.create_remapped_segmentation(out_data_raw, out_data)
+
+    def create_remapped_segmentation(self, out_data_raw: InstanceData, out_data: InstanceData):
+        # Create the remapping dictionary to reorder output labels so they will be picked up in the correct order by DicomSeg
```
LennyN95: 'so they will be picked up in the correct order by DicomSeg' -> 'to match with the annotation in the roi meta parameter', right?

models/gc_spider_baseline/meta.json
LennyN95: `sur- face`

models/gc_spider_baseline/meta.json
LennyN95: How is this algorithm able to run on CT? 
I cannot find any mention of CT in the training or evaluation.

LennyN95 (changes requested):

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
LennyN95: The new structure looks great!

models/gc_spider_baseline/meta.json
silvandeleemput: This was indeed unclear, we clarified this part.

models/gc_spider_baseline/meta.json
silvandeleemput: We changed this to focus on the model instead.

models/gc_spider_baseline/meta.json
silvandeleemput: We have clarified and extended this part.

models/gc_spider_baseline/meta.json
silvandeleemput: We have adjusted the text and added a table with the evaluation results.

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
silvandeleemput: yes, that's correct. I'll rephrase it.

models/gc_spider_baseline/meta.json
silvandeleemput: The original IIS model by Nikolas supported both CT and MRI. However, for the SPIDER challenge the original IIS algorithm by Nikolas was extended with additional labels (discs and spinal canal) and retrained on only MRI data. Therefore, we cannot make any guarantees that the current model works for CT data. 

I realize that I was not fully aware of this fact before and have added the option to run it on CT anyway, but I'll add a big disclaimer in the meta.json.

models/gc_spider_baseline/meta.json
silvandeleemput: We changed this text to be more about the intended use.

models/gc_spider_baseline/utils/SpiderBaselineRunner.py
silvandeleemput: This parameter only matters when the input images are not presented in the expected orientation, i.e. the bottom of the image is expected to start with the lumbar spine, and move upwards to the cervical spine. If this is in the reverse order, you can set this parameter to False to try and match the data instead of flipping the imaging data. So yes the algorithm will produce a different output and we have added a better explanation here.

silvandeleemput:
[output.zip](https://github.com/MHubAI/models/files/14976834/output.zip)

silvandeleemput:
/test

```yaml
sample:
  idc_version: 17.0
  data:
    - SeriesInstanceUID: 1.3.6.1.4.1.14519.5.2.1.1600.1202.719041608264172429762267513295
      aws_url: s3://idc-open-data/6417b84a-f531-4c4a-a47d-3a887bf00f2f/*
      path: dicom

reference:
  url: https://github.com/MHubAI/models/files/14976834/output.zip
```

---

<details><summary>Test Results (24.04.18_22.31.12_E3krCTzQUg)</summary>

```yaml
id: cc8604b9-7b1d-4085-a257-518c035608d9
date: '2024-04-18 22:41:35'
checked_files:
- file: spider_baseline_vertebrae_segmentation_raw.mha
  path: /app/test/src/1.3.6.1.4.1.14519.5.2.1.1600.1202.719041608264172429762267513295/spider_baseline_vertebrae_segmentation_raw.mha
  checks:
  - checker: ImageFileCheck
    notes:
    - label: Data Type
      info: uint16
    - label: Dice Score
      info: 0.999989
- file: spider_baseline_vertebrae.seg.dcm
  path: /app/test/src/1.3.6.1.4.1.14519.5.2.1.1600.1202.719041608264172429762267513295/spider_baseline_vertebrae.seg.dcm
  checks:
  - checker: DicomsegContentCheck
    notes:
    - label: Segment Count
      description: The number of segments identified in the inspected dicomseg file.
      info: 17
- file: spider_baseline_vertebrae.seg.mha
  path: /app/test/src/1.3.6.1.4.1.14519.5.2.1.1600.1202.719041608264172429762267513295/spider_baseline_vertebrae.seg.mha
  checks:
  - checker: ImageFileCheck
    notes:
    - label: Data Type
      info: uint16
    - label: Dice Score
      info: 0.999989
summary:
  files_missing: 0
  files_extra: 0
  checks:
    ImageFileCheck:
      files: 2
    DicomsegContentCheck:
      files: 1
conclusion: true
```

</details>

silvandeleemput:
/review

The meta.json file has been updated and the comments have been addressed. For the provided test file it should be noted that I was unable to find a single case with all the annotated labels of the spine. Typically only a part of the spine is scanned using MR, i.e. the lumbar, thoracic, or cervical vertebrae. The example scan from IDC that I provided is of the lumbar spine.

edit - as discussed, this model only functions properly on input images of the lumbar spine, since it was only trained on these. I think the currently provided test image should be sufficient for testing.

models/gc_spider_baseline/meta.json
```diff
@@ -0,0 +1,219 @@
...
+    },
+    "evaluation": {
+      "title": "Evaluation data",
+      "text": "The test set consists of 87 lumbar MRI images from 39 patients, which includes the fully manually annotated images of 15 patients, randomly selected from the set of fully annotated dataset, see Training data.",
+      "references": [],
+      "tables": []
+    },
+    "training": {
+      "title": "Training data",
+      "text": "For training a large, publicly available, multi-center, lumbar spine MRI dataset was used. It includes reference segmentations of vertebrae, intervertebral discs (IVDs), and spinal canal. The dataset includes 447 sagittal T1 and T2 MRI images from 218 patients with a history of low back pain. For 20 patients from the UMC dataset the images were fully, manually annotated. This model is based on a previously developed model trained on both CT and MRI data [1]. It was extended by also segmenting the IVDs and spinal canal, using MRI only. For training details see Baseline 1 in the associated paper [2].",
```
LennyN95: We should also add the link to the dataset as it's mentioned to be public: 
https://doi.org/10.5281/zenodo.10159290

models/gc_spider_baseline/meta.json
```diff
@@ -0,0 +1,219 @@
+{
+  "id": "aed5a82e-2188-4d50-83f9-45bce09f9daa",
+  "name": "gc_spider_baseline",
+  "title": "SPIDER Baseline algorithm: Iterative Instance Segmentation of the spine",
+  "summary": {
+    "description": "This model segments three anatomical structures in lumbar spine MRI: vertebrae, intervertebral discs (IVDs), and spinal canal. Each vertebrae and IVD is labeled according to standard naming convention.",
+    "inputs": [
+      {
+        "label": "Sagittal Lumbar Spine T1 T2 MRI",
+        "description": "Sagittal T1 or T2 weighted MRI of the lumbar spine. It is important to provide the input image in sagittal orientation for proper functioning of the model.",
```
LennyN95, line 10: > It is important to provide the input image in sagittal orientation for proper functioning of the model
This should also go into the intended use section.

models/gc_spider_baseline/meta.json
```diff
@@ -0,0 +1,219 @@
...
+    ],
+    "github": "https://github.com/DIAGNijmegen/SPIDER-Baseline-IIS",
+    "zenodo": "",
+    "colab": "",
+    "slicer": false
+  },
+  "info": {
+    "use": {
+      "title": "Intended use",
+      "text": "This model is intended to be used with T1 or T2 weighted MRI images of the lumbar spine in sagittal orientation only. Ensure that the lumbar spine is visible in the input MRI images to ensure reliable results. This algorithm is available on Grand Challenge [1] and is used as baseline for the SPIDER challenge [2].",
```
LennyN95: > Ensure that the lumbar spine is visible in the input MRI images to ensure reliable results

The consequence (unreliable/unexpected results when the lumbar spine is absent) should also be included in the Limitations section.

The limitations section aims to give the reader fast access to information about what the model cannot.

LennyN95 (changes requested):

models/gc_spider_baseline/meta.json
miriam-groeneveld: It was mentioned in the Intended use section, but I've made it more explicit.

models/gc_spider_baseline/meta.json
miriam-groeneveld: Added it to the limitations

models/gc_spider_baseline/meta.json
miriam-groeneveld: Done

LennyN95 (approved):
Looking good & tests passed.

PR #79, opened 2024-05-01
BAMF FDG-Avid Breast Tumor

jithenece:


vanossj:
@LennyN95 What is the right way to have multi modality inputs into the `NNUnetRunner`? We made a custom runner, but it doesn't feel like that should be required.

LennyN95:
@vanossj if it's an nnunet that takes e.g. multiple modalities interchangeably but only one input at a time, then you could use the `|` operator like in `nifti:mod=mr|ct`.
However, if you have multiple inputs (as your implementation suggests), the current version of the NNUnet runner indeed isn't capable of handling this. We are planning to update this (however, on a lower prior). So, in this case, a custom module (you can, of course, start from our implementation and simply make the desired modifications) would be the fastest.

vanossj:
@LennyN95 Could you add a roi for FDG-avid tumor? (NCIt C168968). if you have an alternative sct code that would work too

LennyN95:
@jithenece @vanossj As per your suggestion, I added `FDG_AVID_TUMOR` to SegDB.

jithenece:
@LennyN95 could you add `BREAST` to segdb as it is required for generating `roi=BREAST+FDG_AVID_TUMOR`.

LennyN95:
@jithenece done, added `BREAST` as new segmentation to SegDB.

jithenece:
I tried to test this but seems to fail. Does this require updating base image mhubai/base:latest.
`sha256:d72b350c7f98e5f2460297d6059a70f967091c9ed11728ab3bbc7f9b32b65d2e`

> @jithenece done, added `BREAST` as new segmentation to SegDB.

jithenece:
closing this as this will be done in another [pull request ](https://github.com/MHubAI/models/pull/87)

PR #82, opened 2024-06-07
[FEAT]: BAMF nnUNet Breast MR Model

rahul99:


rahul99:
@LennyN95 , @fedorov : Hey guys, we would like to add another segmentation model to MHub that segments Breast, FGT, and Breast tumor. Can you please make an entry to SegDB for breast segmentation relation ROIs?
```
BREAST
FGT
BREAST_TUMOR
```

rahul99:
@LennyN95 , it looks like MHubIO only support nnUNet-v1 models at the moment? I am referring to this code here: https://github.com/MHubAI/mhubio/blob/main/mhubio/modules/runner/NNUnetRunner.py#L21

The breast segmentation model was trained on nNUNet-v2. Let me know how to proceed with this. cc. @vanossj

LennyN95:
Hi @rahul99, this sounds like an interesting model and another great addition to the MHub model repository!

- Can you prepare a [DCMQI style JSON](http://qiicr.org/dcmqi/#/seg) or provide the codes in an alternative format? I'll then add them to our SegDB. I assume we only need to add `BREAST` and `FGT`, you then can use `BREAST+TUMOR` for the breast neoplasm.
- Our internal NNUnetRunner, indeed, is based on the v1. I have a v2 version in the works. However, you can implement the model and provide your own NNUnetV2Runner.

rahul99:
```json
{
  "ContentCreatorName": "BAMFHealth^AI",
  "ClinicalTrialSeriesID": "Session1",
  "ClinicalTrialTimePointID": "1",
  "SeriesDescription": "AIMI breast-tumor AI segmentation",
  "SeriesNumber": "300",
  "InstanceNumber": "1",
  "BodyPartExamined": "BREAST",
  "segmentAttributes": [
    [
      {
        "labelID": 1,
        "SegmentDescription": "Breast",
        "SegmentAlgorithmType": "AUTOMATIC",
        "SegmentAlgorithmName": "BAMF-Breast-Tumor-MRI",
        "SegmentedPropertyCategoryCodeSequence": {
          "CodeValue": "123037004",
          "CodingSchemeDesignator": "SCT",
          "CodeMeaning": "Anatomical Structure"
        },
        "SegmentedPropertyTypeCodeSequence": {
          "CodeValue": "76752008",
          "CodingSchemeDesignator": "SCT",
          "CodeMeaning": "Breast"
        },
        "recommendedDisplayRGBValue": [56, 125, 255]
      },
      {
        "labelID": 2,
        "SegmentDescription": "Fibroglandular Tissue",
        "SegmentAlgorithmType": "AUTOMATIC",
        "SegmentAlgorithmName": "BAMF-Breast-Tumor-MRI",
        "AnatomicRegionSequence": {
          "CodeValue": "76752008",
          "CodingSchemeDesignator": "SCT",
          "CodeMeaning": "Breast"
        },
        "AnatomicRegionModifierSequence": {
          "CodeValue": "51440002",
          "CodingSchemeDesignator": "SCT",
          "CodeMeaning": "Right and left"
        },
        "SegmentedPropertyCategoryCodeSequence": {
          "CodeValue": "C35869",
          "CodingSchemeDesignator": "NCIt",
          "CodeMeaning": "Radiologic Finding"
        },
        "SegmentedPropertyTypeCodeSequence": {
          "CodeValue": "C114467",
          "CodingSchemeDesignator": "NCIt",
          "CodeMeaning": "Breast Fibroglandular Tissue"
        },
        "recommendedDisplayRGBValue": [128, 174, 128]
      },
      {
        "labelID": 3,
        "SegmentDescription": "Structural Tumor",
        "SegmentAlgorithmType": "AUTOMATIC",
        "SegmentAlgorithmName": "BAMF-Breast-Tumor-MRI",
        "AnatomicRegionSequence": {
          "CodeValue": "76752008",
          "CodingSchemeDesignator": "SCT",
          "CodeMeaning": "Breast"
        },
        "AnatomicRegionModifierSequence": {
          "CodeValue": "51440002",
          "CodingSchemeDesignator": "SCT",
          "CodeMeaning": "Right and left"
        },
        "SegmentedPropertyCategoryCodeSequence": {
          "CodeValue": "49755003",
          "CodingSchemeDesignator": "SCT",
          "CodeMeaning": "Morphologically Altered Structure"
        },
        "SegmentedPropertyTypeCodeSequence": {
          "CodeValue": "4147007",
          "CodingSchemeDesignator": "SCT",
          "CodeMeaning": "Mass"
        },
        "recommendedDisplayRGBValue": [144, 238, 144]
      }
    ]
  ],
  "ContentLabel": "SEGMENTATION",
  "ContentDescription": "Image segmentation",
  "ClinicalTrialCoordinatingCenterName": "dcmqi"
}
```

rahul99:
@LennyN95 , Hey Leo, I have posted the json containing code for ROIs above. Please have a look and let me know if I missed something. I have also added NNUnetRunnerV2. Its an easy extension of the NNUnetRunner that you wrote so beautifully.

Its a slightly different PR as it contains two models so there might be something I probably missed. Please review and let me know :)

LennyN95:
@rahul99, this looks great so far! I noticed that you use `Mass (4147007)` as a type for the breast tumor. Is there any reason why you choose not to use `neoplasm (86049000)` here? We can add a new tumor class if required, but I'd like to understand the background first.

vanossj:
we can switch it to neoplasm

rahul99:
@LennyN95 , since we did not have the ROI entries in SegDB, I was testing the container by specifying other ROIs that are available in SegDB. Since the process outputs three ROIs and I see that entry for `BREAST` label is available already, I tried two random ROIs. Unfortunately, the flow complained about `BREAST` label too. Please can you look into this?

The flow worked great when I passed three random ROIs that are available in SegDB (Except `BREAST` label)

I think refreshing the base image might solve this?

rahul99:
@LennyN95 , a small query on the NNUnetRunner as well. In this line:
https://github.com/MHubAI/mhubio/blob/main/mhubio/modules/runner/NNUnetRunner.py#L165

Comment says add symlink to input dir, but we add symlink to output dir. Also, I don't think that's needed? We have already organised the data into appropriate folder structure and nnunet predict function does not use it. Let me know what your thoughts are on this?

rahul99:
@LennyN95 , Hey Leo, do you have any updates on the SegDB entries for Breast and FGT?

LennyN95:
Hey @rahul99, 

I was looking into this and found that it'd be good to discuss briefly with @dclunie if available at PW. 
I could add `FGT` as specified, however, that you only be used in the context of breast, thus `BREAST+FGT` would be the only valid usage of FGT, `LUNG+FGT` for instance wouldn't make any sense. We currently have no options to define such bindings (and the question is weather we have to). One solution could be to make `FGT` more verbose, e.g., by naming the ID `BREAST_FGT`, however, then the usage is `BREAST+BREAST_FGT`. I personally don't mind the redundancy in this specific situation but it might cause confusion since `BREAST_FGT` should not be used standalone.

In the meanwhile, I added an option to specify custom triplets and segments in the config file that will be available as SegDB ID's at runtime.

You can either register [custom triplets](https://github.com/MHubAI/SegDB/blob/6b2db45ecdc711c5dee93990b73594ccd6a77d9a/segdb/classes/Triplet.py#L63) and [custom segments](https://github.com/MHubAI/SegDB/blob/6b2db45ecdc711c5dee93990b73594ccd6a77d9a/segdb/classes/Segment.py#L40) in your runner module class (outside of the class definition) or define them in the config file as explained [here](https://github.com/MHubAI/documentation/blob/main/documentation/mhubio/the_mhubio_config_file.md#segdb-customization).

*Note, that custom triplets and segments should mainly be used to simplify development and to accommodate the time until a segdb update propagates into a new base image. However, in special cases where we won't add a segment to our universal SegDB this is a transparent way to define custom segments.*

dclunie:
I am not in Boston for PW41 but am available online; FYI, since the subject of codes for FGT came up, see DICOM [CP 2226](https://www.dclunie.com/dicom-status/status.html#CP2226) (currently out for ballot).

fedorov:
@LennyN95 since you introduced that capability of registering custom triplets, it might be good to add some basic validation checks. Coding scheme designator must come from the list in the standard: https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_8.html, or must start with the "99" prefix.

LennyN95:
Good point @fedorov. The default is indeed 99segdb, is hardxoding the list and validating against it a valuable solution? I assume this doesn't change frequently if at all?

fedorov:
> The default is indeed 99segdb

I believe the recommendation is to have all caps, so it would be 99SEGDB. But private coding scheme should be absolutely last resort!

> I assume this doesn't change frequently if at all?

Yes, they don't. But a basic check (ie if it does not start from 99 it must come from the defined list) would prevent someone from using, for example, an invalid value "LUNGMASK" for coding scheme designator.

LennyN95:
@fedorov I implemented it all uppercase, see https://github.com/MHubAI/SegDB/blob/6b2db45ecdc711c5dee93990b73594ccd6a77d9a/segdb/classes/Triplet.py#L63C105-L63C112 ;)

Just to make clear, I asked if this changes frequently to get an idea if hardcoding the list to check against is a valuable solution, I assume yes then. This is a simple check we can easily include! Note that in general, this is more thought to be used during development to have a quick placeholder and then to be resolved during the review process.

rahul99:
Hey @LennyN95,  I have added custom SegDB keys in the config. Can you please help me review and see it the settings look all right? I still got the key error when I tried to run the container locally:

```
Invalid segment ids: FGT, BREAST+TUMOR
```

Can you please advice if I missed anything? I am hoping to close this soon, please help with the segdb config and then I will post the test case and upload segmentation images

LennyN95:
@rahul99 the correct keyword is `segments` not `segmentations`, this was outdated in the documentation. Sorry for the confusion, I updated our docs accordingly. Let me know if it works with the updated config.

rahul99:
/test

```yaml
sample:
  idc_version: 17.0
  data:
  - SeriesInstanceUID: 1.3.6.1.4.1.14519.5.2.1.57828200249886597927204318690632931889
    aws_url: s3://idc-open-data/bf0fc92d-e708-468c-af90-0c02329f636d/*
    path: dicom

reference:
  url: https://www.dropbox.com/scl/fi/o2b26zrb20e080rcc0z5d/bamf_nnunet_mr_breast.zip?rlkey=8m434go1u7zthemx0w245pors&st=wetbolof&dl=0
```

rahul99:
@LennyN95 , the issue was also around SegDB logic to fetch ROI keys. I believe my teammate (Jith) got it in touch with you about that? For now, everything works all right after adding a custom logic to get ROI keys inside the transformation script.

I have added a test case, and as before, uploaded the segmentation in dropbox. Let me know how it goes.

Note: We have used a test case from a different dataset that is not part of training or testing the model

LennyN95:
Thank you @rahul99. The custom ROI should always have workes, only the keyword to use in the config file has changed and that change hasn't been reflected in the documentation. 

Can you remove the custom logic and test with the updated config as mentioned above? Then we can proceed with testing and pushing the model. Thank you!

rahul99:
Hey @LennyN95 , moving the segDB custom triplets and segments to config still fails with following error trace:

```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.8/dist-packages/mhubio/run.py", line 613, in <module>
    run(config_file)
  File "/usr/local/lib/python3.8/dist-packages/mhubio/run.py", line 374, in run
    config = Config(config_file=config_file, config=config_data)
  File "/usr/local/lib/python3.8/dist-packages/mhubio/core/Config.py", line 143, in __init__
    for seg_id, seg_data in self._config['segdb']['segments']:
ValueError: too many values to unpack (expected 2)
```

I think there's a small bug. If multiple segment are specified, `self._config['segdb']['segments']` would return a dict object instead of values. So the following tweak?

```
for segment in self._config['segdb']['segments']:
    for seg_id, seg_data in x['segdb']['segments'][segment].items():
        ...
```

cc. @vanossj

LennyN95:
Oh indeed, if I see correctly, then all it takes is adding the missing “.items()” here https://github.com/MHubAI/mhubio/blob/main/mhubio/core/Config.py#L143C41-L143C75. This got lost somehow. I'll be on vacation until the 11th, I will update you once the fix is deployed and the newest base image available!

rahul99:
@LennyN95 , enjoy your vacation. Drop us a message when a new base image is published and I will test it on my end. However, I think the automated test would trigger here anyway!

rahul99:
[@LennyN95] Hey Leo, hope you had a good vacation. Just a quick reminder on this pending action item. Looking forward to getting this PR over the line. Thanks!

LennyN95:
Hey @rahul99, I implemented the fix alongside some other improvements and uploaded a new base image. Please make sure you have the latest version of the base image (you can run `docker pull mhubai/base` and check the version by running `docker run --rm -t --entrypoint mhub.version mhubai/base | grep mhubio`).

rahul99:
Hey Leo, the codebase works with the newer base image. Since there is no code change, will the test instance be picked up already? I am reposting the test instance just in case. Let me know how it goes. Thanks!

rahul99:
/test

```yaml
sample:
  idc_version: 17.0
  data:
  - SeriesInstanceUID: 1.3.6.1.4.1.14519.5.2.1.57828200249886597927204318690632931889
    aws_url: s3://idc-open-data/bf0fc92d-e708-468c-af90-0c02329f636d/*
    path: dicom

reference:
  url: https://www.dropbox.com/scl/fi/o2b26zrb20e080rcc0z5d/bamf_nnunet_mr_breast.zip?rlkey=8m434go1u7zthemx0w245pors&st=wetbolof&dl=0
```

---

<details><summary>Test Results (24.08.29_15.15.30_AkoraywFNq)</summary>

```yaml
id: 1a76a990-9c3a-46d7-a38c-7b5f9eb54fe4
date: '2024-08-29 16:06:16'
missing_files:
- 1.3.6.1.4.1.14519.5.2.1.57828200249886597927204318690632931889/bamf_nnunet_mr_breast.seg.dcm
summary:
  files_missing: 1
  files_extra: 0
  checks: {}
conclusion: false
```

</details>

models/bamf_nnunet_mr_breast/utils/NNUnetRunnerV2.py
```diff
@@ -0,0 +1,210 @@
...
+nnunet_dataset_name_regex = r"Dataset[0-9]{3}_[a-zA-Z0-9_]+"
+
+@IO.ConfigInput('in_data', 'nifti:mod=mr', the="input data to run nnunet on")
+@IO.Config('nnunet_dataset', str, None, the='nnunet dataset name')
+@IO.Config('nnunet_config', str, None, the='nnunet model name (2d, 3d_lowres, 3d_fullres, 3d_cascade_fullres)')
+@IO.Config('folds', int, None, the='number of folds to run nnunet on')
+@IO.Config('use_tta', bool, True, the='flag to enable test time augmentation')
+@IO.Config('export_prob_maps', bool, False, the='flag to export probability maps')
+@IO.Config('prob_map_segments', list, [], the='segment labels for probability maps')
+@IO.Config('roi', str, None, the='roi or comma separated list of roi the nnunet segments')
```
LennyN95: Same as in https://github.com/MHubAI/models/pull/84#discussion_r1722431622.

models/bamf_nnunet_mr_breast/utils/NNUnetRunnerV2.py
```diff
@@ -0,0 +1,210 @@
...
+            model_output_folder : required - path to the folder where the inferred segmentation masks should be stored.
+            ref_file            : required - InstanceData object of the generated segmentation mask used as reference file.
+            output_dtype        : optional - output data type. Data type float16 is not supported by the NRRD standard,
+                                            so the choice should be between uint8, uint16 or float32. 
+            structure_list      : optional - list of the structures whose probability maps are stored in the 
+                                            first channel of the `.npz` file (output from the nnU-Net pipeline
+                                            when `export_prob_maps` is set to True). 
+        Outputs:
+            This function [...]
+        """
```
LennyN95: Do we need this in this specific implementation? If not, I suggest we remove it to keep all custom modules as specific and simple to use for the end-user.

models/bamf_nnunet_mr_breast/meta.json
```diff
@@ -0,0 +1,134 @@
...
+      ],
+      "outputs": [
+        {
+          "label": "Segmentation",
+          "type": "Segmentation",
+          "description": "Segmentation for breast, fgt, and tumor",
+          "classes": [
+            "BREAST",
+            "FGT",
+            "BREAST+TUMOR"
```
LennyN95: `BREAST+ BREAST_CARCINOMA`?

models/bamf_nnunet_mr_breast/config/default.yml
```diff
@@ -0,0 +1,67 @@
...
+      type: T_BREAST_CARCINOMA_MASS
+      modifier: M_RIGHT_AND_LEFT
+      color: [144, 238, 144]
+
+execute:
+- DicomImporter
+- NiftiConverter
+- module: NNUnetRunnerV2
+  nnunet_dataset: Dataset009_Breast
+  roi: BREAST,FGT
```
LennyN95: I think it should be `BREAST+FGT` (although FGT is always breast-specific - what do you think?).

LennyN95: If changed, changes also need to be applied in the meta.json output section.

models/bamf_nnunet_mr_breast/meta.json
```diff
@@ -0,0 +1,134 @@
...
+    "title": "BAMF MR Breast Segmentation (nnU-Net)",
+    "summary": {
+      "description": "An nnU-Net based model to segment breast from MR scans",
+      "inputs": [
+        {
+          "label": "Input Image",
+          "description": "The MR scan of a patient.",
+          "format": "DICOM",
+          "modality": "MR",
+          "bodypartexamined": "BREAST,FGT,BREAST+TUMOR",
```
LennyN95: These are not valid labels under BPE (see https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_L.html#chapter_L), so `BREAST` would be correct here.

models/bamf_nnunet_mr_breast/utils/BreastPostProcessor.py
```diff
@@ -0,0 +1,101 @@
...
+            tumor_seg = np.copy(tumor_seg),
+            mr_path = in_mr_data.abspath
+            )
+        process_dir = self.config.data.requestTempDir(label="nnunet-breast-processor")
+        process_file = os.path.join(process_dir, f'bamf_nnunet_mr_breast.nii.gz')
+        sitk.WriteImage(
+            output_seg,
+            process_file,
+        )
+        shutil.copyfile(process_file, out_data.abspath)
```
LennyN95: Instead of declaring an extra process file, you can also use `out_data.abspath` directly:

```Python
sitk.WriteImage(
    output_seg,
    out_data.abspath
)
```

LennyN95 (changes requested):
Great implementation. Some comments and minor change-requests below.

models/bamf_nnunet_mr_breast/utils/NNUnetRunnerV2.py
rahul99: Thanks, I have removed probability map related functions because we are not doing that.

models/bamf_nnunet_mr_breast/utils/NNUnetRunnerV2.py
rahul99: Correct, we don't need that. Removing it

rahul99:
Hey @LennyN95 , above comments are resolved. Let me know if you have more suggestions you would like me to include. Thanks!

models/bamf_nnunet_mr_breast/config/default.yml
```diff
@@ -31,7 +31,7 @@ execute:
 - NiftiConverter
 - module: NNUnetRunnerV2
   nnunet_dataset: Dataset009_Breast
-  roi: BREAST,FGT
+  roi: BREAST+FGT
```
LennyN95: Just to clarify, when it is annotated like this, it means that the model produces a single foreground structure (label id 1) which is the FGT - correct?

models/bamf_nnunet_mr_breast/meta.json
```diff
@@ -20,11 +20,10 @@
         {
           "label": "Segmentation",
           "type": "Segmentation",
-          "description": "Segmentation for breast, fgt, and tumor",
+          "description": "Segmentation for breast, fgt, and breast carcinoma",
           "classes": [
-            "BREAST",
```
LennyN95: Does the model generate a delineation of the breast too or "only" FGT and TUMOR then?

models/bamf_nnunet_mr_breast/config/default.yml
```diff
@@ -0,0 +1,67 @@
...
+    source_dir: input_data
+    import_dir: sorted_data
+    sort_data: true
+    meta: 
+      mod: '%Modality'
+
+  NNUnetRunnerV2:
+    in_data: nifti:mod=mr
+    nnunet_config: 3d_fullres
+    export_prob_maps: False
```
LennyN95: Now can be removed.

models/bamf_nnunet_mr_breast/config/default.yml
```diff
@@ -0,0 +1,67 @@
...
+  DicomImporter:
+    source_dir: input_data
+    import_dir: sorted_data
+    sort_data: true
+    meta: 
+      mod: '%Modality'
+
+  NNUnetRunnerV2:
+    in_data: nifti:mod=mr
+    nnunet_config: 3d_fullres
```
LennyN95: Should this parameter be configurable? I believe it's highly specific to the trained model and shouldn't be changed by the user.

models/bamf_nnunet_mr_breast/utils/NNUnetRunnerV2.py
```diff
@@ -0,0 +1,108 @@
...
+import os, subprocess, shutil
+import SimpleITK as sitk, numpy as np
+from mhubio.core import Module, Instance, InstanceData, DataType, FileType, IO
+
+
+nnunet_dataset_name_regex = r"Dataset[0-9]{3}_[a-zA-Z0-9_]+"
+
+@IO.ConfigInput('in_data', 'nifti:mod=mr', the="input data to run nnunet on")
+@IO.Config('nnunet_dataset', str, None, the='nnunet dataset name')
+@IO.Config('nnunet_config', str, None, the='nnunet model name (2d, 3d_lowres, 3d_fullres, 3d_cascade_fullres)')
```
LennyN95: see above;

models/bamf_nnunet_mr_breast/utils/NNUnetRunnerV2.py
```diff
@@ -0,0 +1,108 @@
...
+@IO.Config('folds', int, None, the='number of folds to run nnunet on')
+@IO.Config('roi', str, None, the='roi or comma separated list of roi the nnunet segments')
+class NNUnetRunnerV2(Module):
+
+    nnunet_dataset: str
+    nnunet_config: str
```
LennyN95: see above;

models/bamf_nnunet_mr_breast/utils/NNUnetRunnerV2.py
```diff
@@ -0,0 +1,108 @@
...
+    @IO.Input("in_data", the="input data to run nnunet on")
+    @IO.Output("out_data", 'VOLUME_001.nii.gz', 'nifti:mod=seg:model=nnunet', data='in_data', the="output data from nnunet")
+    def task(self, instance: Instance, in_data: InstanceData, out_data: InstanceData) -> None:
+        
+        # get the nnunet model to run
+        self.v("Running nnUNet_predict.")
+        self.v(f" > dataset:     {self.nnunet_dataset}")
+        self.v(f" > config:      {self.nnunet_config}")
+        self.v(f" > input data:  {in_data.abspath}")
+        self.v(f" > output data: {out_data.abspath}")
```
LennyN95: Any debug output should indicate that this is a NNUnet v2 (custom) implementation. At least the v2 should be mentioned (in contrast to v1).

LennyN95 (changes requested):

models/bamf_nnunet_mr_breast/meta.json
rahul99: Hey Leo, sorry for the confusion. Yes the model generates a delineation of breast, produces a total of three output labels

models/bamf_nnunet_mr_breast/config/default.yml
rahul99: We do have models in the pipeline where we tried cascade model too. So I left that option there for the future. But as far as this specific model goes, it only has `3d_fullres`. I have removed it from the config for this iteration

LennyN95: Okay, either is fine in that case. @rahul99, I now implemented an (experimental) routine to host reusable modules in a separate mhubio extension repository which can then be reused across mhubio modules. Since you use some custom modules for a variety of models, this could be an interesting use-case (like a BAMF-utility-collection). The injection of such mhubio collections is explained [here](https://github.com/MHubAI/documentation/blob/main/documentation/mhub_models/the_mhub_dockerfile.md#install-additional-mhub-io-collections). Since we don't have documentation on how to write such collection repositories, we can schedule a call to demonstrate the current approach if that sounds interesting to you.

rahul99: @vanossj , fyi

vanossj:
@rahul99, lets switch the FGT code to [SCT 1342375003](https://browser.ihtsdotools.org/?perspective=full&conceptId1=1342375003&edition=MAIN/2024-08-01&release=&languages=en&latestRedirect=false). That way we can keep codes in snowmed as much as possible.

so change metadata for label 2 to:
```json
{
  "labelID": 2,
  "SegmentDescription": "Fibroglandular Tissue",
  "SegmentAlgorithmType": "AUTOMATIC",
  "SegmentAlgorithmName": "BAMF-Breast-Tumor-MRI",
  "AnatomicRegionSequence": {
    "CodeValue": "76752008",
    "CodingSchemeDesignator": "SCT",
    "CodeMeaning": "Breast"
  },
  "AnatomicRegionModifierSequence": {
    "CodeValue": "51440002",
    "CodingSchemeDesignator": "SCT",
    "CodeMeaning": "Right and left"
  },
  "SegmentedPropertyCategoryCodeSequence": {
    "CodeValue": "123037004",
    "CodingSchemeDesignator": "SCT",
    "CodeMeaning": "Anatomical Structure"
  },
  "SegmentedPropertyTypeCodeSequence": {
    "CodeValue": "1342375003",
    "CodingSchemeDesignator": "SCT",
    "CodeMeaning": "Structure of fibroglandular tissue of breast"
  },
  "recommendedDisplayRGBValue": [128, 174, 128]
}
```

models/bamf_nnunet_mr_breast/config/default.yml
vanossj: The goal is to get the AIMI PRs finished next week. I would rather not start implementing an experimental feature now.

vanossj (commented):
`category: C_RADIOLOGIC_FINDING` should change too

models/bamf_nnunet_mr_breast/config/default.yml
```diff
@@ -0,0 +1,65 @@
...
+  triplets:
+    T_BREAST_FGT_STRUCTURE:
+      code: 1342375003
+      meaning: Structure of fibroglandular tissue of breast
+      scheme_designator: SCT
+    T_BREAST_CARCINOMA_MASS:
+      code: 4147007
+      meaning: Mass
+      scheme_designator: SCT
+  segments:
```
LennyN95, line 16: SegDB now has IDs for `BREAST_FIBROGLANDULAR_TISSUE` and `BREAST_CARCINOMA`.

models/bamf_nnunet_mr_breast/config/default.yml
```diff
@@ -0,0 +1,65 @@
...
+      name: FGT
+      category: C_RADIOLOGIC_FINDING
+      type: T_BREAST_FGT_STRUCTURE
+      color: [128, 174, 128]
+    BREAST_CARCINOMA:
+      name: Breast_Carcinoma
+      category: C_MORPHOLOGICALLY_ABNORMAL_STRUCTURE
+      type: T_BREAST_CARCINOMA_MASS
+      modifier: M_RIGHT_AND_LEFT
+      color: [144, 238, 144]
```
LennyN95: SegDB now has IDs for `BREAST_FIBROGLANDULAR_TISSUE` and `BREAST_CARCINOMA`.

LennyN95 (commented):
Run mhub.update inside of your image (during development) or re-build with the latest base image `docker pull mhubai/base`.

rahul99:
@LennyN95 , thanks for making breast related entries available on SegDB. I've rolled back the config :)

On standby if any more changes required. Thanks!

LennyN95:
@rahul99 I saw you use the `BREAST+BREAST_FIBROGLANDULAR_TISSUE` roi, however, since we switched to the SCT codes as suggested by @vanossj, this code represents a tissue now (rather than a radiological finding) and thus the `BREAST+` context can (and should) be omitted.

rahul99:
> @rahul99 I saw you use the `BREAST+BREAST_FIBROGLANDULAR_TISSUE` roi, however, since we switched to the SCT codes as suggested by @vanossj, this code represents a tissue now (rather than a radiological finding) and thus the `BREAST+` context can (and should) be omitted.

I have made the change. Thanks for educating me on this as well. :)

rahul99:
[bamf_nnunet_mr_breast.zip](https://github.com/user-attachments/files/16875754/bamf_nnunet_mr_breast.zip)

rahul99:
@LennyN95 , following your [comment on the other PR](https://github.com/MHubAI/models/pull/87#issuecomment-2310719504), I have uploaded the output segmentation file above. Thanks!

rahul99:
/test

```yaml
sample:
  idc_version: 17.0
  data:
  - SeriesInstanceUID: 1.3.6.1.4.1.14519.5.2.1.57828200249886597927204318690632931889
    aws_url: s3://idc-open-data/bf0fc92d-e708-468c-af90-0c02329f636d/*
    path: dicom

reference:
  url: https://www.dropbox.com/scl/fi/o2b26zrb20e080rcc0z5d/bamf_nnunet_mr_breast.zip?rlkey=8m434go1u7zthemx0w245pors&st=wetbolof&dl=0
```

[bamf_nnunet_mr_breast.zip](https://github.com/user-attachments/files/16894948/bamf_nnunet_mr_breast.zip)

LennyN95:
Please note, we **updated our base image**. All mhub dependencies are now installed in a virtual environment under `/app/.venv` running `Python 3.11`. Python, virtual environment and dependencies are now managed with [uv](https://github.com/astral-sh/uv). If required, you can create custom virtual environments, e.g., `uv venv -p 3.8 .venv38` and use `uv pip install -p .venv38 packge-name` to install dependencies and `uv run -p .venv3.8 python script.py` to run a python script. 

We also **simplified our test routine**. Sample and reference data now have to be uploaded to Zenodo and provided in a mhub.tom file at the project root. The process how to create and provide these sample data is explained in the updated [testing phase article](https://github.com/MHubAI/documentation/blob/main/documentation/mhub_contribution/testing_phase.md) of our documentation. Under [doi.org/10.5281/zenodo.13785615](https://doi.org/10.5281/zenodo.13785615) we provide sample data as a reference.

LennyN95:
@rahul99 thx for adapting to the new test routine! To make the test pass, you can reference the download link to the zip file (https://zenodo.org/records/13843612/files/bamf_nnunet_mr_breast.test.zip?download=1).

rahul99:
@LennyN95 , thanks for pointing it out. Let me know if you need more information from us. Looking forward to getting this PR over the line

rahul99:
/test

```yaml
sample:
  idc_version: 17.0
  data:
  - SeriesInstanceUID: 1.3.6.1.4.1.14519.5.2.1.57828200249886597927204318690632931889
    aws_url: s3://idc-open-data/bf0fc92d-e708-468c-af90-0c02329f636d/*
    path: dicom

reference:
  url: https://zenodo.org/records/13843612/files/bamf_nnunet_mr_breast.test.zip?download=1
```

rahul99:
@LennyN95 , just circling back to see if you have an update for us. Thanks!

LennyN95:
Since this model relies on NNUnet V2, I paused the submission process until https://github.com/MIC-DKFZ/nnUNet/issues/2556 is resolved. This will lead to an improved UX for MHub models and we're actively working on this now. Let me know if you are running against any deadlines and feel free to share your opinion / feedback in the related issue. Thx @rahul99!

PR #85, opened 2024-06-27
[PW41] Add body part regression base

surajpaib:

PR #98, opened 2024-09-11
Integrate body part regression

surajpaib:
Switching base branch.

Original PR: https://github.com/MHubAI/models/pull/85


LennyN95:
Please note, we **updated our base image**. All mhub dependencies are now installed in a virtual environment under `/app/.venv` running `Python 3.11`. Python, virtual environment and dependencies are now managed with [uv](https://github.com/astral-sh/uv). If required, you can create custom virtual environments, e.g., `uv venv -p 3.8 .venv38` and use `uv pip install -p .venv38 packge-name` to install dependencies and `uv run -p .venv3.8 python script.py` to run a python script. 

We also **simplified our test routine**. Sample and reference data now have to be uploaded to Zenodo and provided in a mhub.tom file at the project root. The process how to create and provide these sample data is explained in the updated [testing phase article](https://github.com/MHubAI/documentation/blob/main/documentation/mhub_contribution/testing_phase.md) of our documentation. Under [doi.org/10.5281/zenodo.13785615](https://doi.org/10.5281/zenodo.13785615) we provide sample data as a reference.

LennyN95:
If we parse the BPE results in MHub value outputs (using the dedicated IO decorators) the output format can be customized through a report exporter. What do you think?

surajpaib:
@LennyN95 I'll take a look at this! I'm thinking Deepa might having something that we can use here as well as she did BPR -> Structured Report. 

But maybe we get in the first json version of this integrated and then add a workflow. I'll add tests based on the update!

surajpaib:
@LennyN95 I'm running into some challenges with the python environment for body part regression. I will debug this further and see. I think it might be better to use a custom fork with updated dependencies. 

Is there an mhub policy on this?

surajpaib:
I think this is from the new mhub container. Was there an OS upgrade as well? 

I've tried to use Python3.8 - which it worked on earlier but breaks now

LennyN95:
No so far I only updated Python. Have you tried installing bpr into a separate environment?

surajpaib:
I have yes - on two separate machines but both of them have had OS upgrades in the past weeks so I was attributing it to that. But now I am not sure anymore.

surajpaib:
@LennyN95 I've added a test for the default workflow but it seems like the compliance check fails because I haven't added tests for other workflows. Is this intended? Any way I can skip and only add tests for default as before?

LennyN95:
> I've added a test for the default workflow but it seems like the compliance check fails because I haven't added tests for other workflows. Is this intended? 

Yes, we indeed need sample and reference data for every submitted workflow. It looks like the submission currently contains only one default workflow. I recently fixed an issue with the compliance checks. I will close and reopen the PR to rerun the automation. If all compliance checks pass, we can proceed with the testing.

LennyN95:
@surajpaib can you address the following issues before we proceed with the testing of the model:

```
Errors:
- Model folder models/bpreg does not contain a mhub.toml
  https://github.com/MHubAI/documentation/blob/main/documentation/mhub_models/model_folder_structure.md
  https://github.com/MHubAI/documentation/blob/main/documentation/mhub_contribution/testing_phase.md#test-procedure
- Dockerfile does not contain the correct FROM command: FROM mhubai/base:v1
  https://github.com/MHubAI/documentation/blob/main/documentation/mhub_models/the_mhub_dockerfile.md

Warnings:
- Model toml validation skipped because file does not exist.
- Check for test data validation skipped because mhub.toml does not exist.
```

Thank you!

LennyN95:
@surajpaib How is the progress on this one going?

PR #109, opened 2025-05-22
msk_smit_lung_gtv, change num_workers=0 in run_segmentation.py 

locastre:
For data_loader low-shared memory instances. Should not affect runtime.


locastre:
@LennyN95 We made a trivial change to one parameter in the script that runs our segmentation routine. We were running into issues where the routine would "hang" on systems that had low shared memory (`/dev/shm`) configured in the Docker container. By reducing the `num_workers=0` we avoid running into the issue and the routine runs just as quickly.

